Issue #1076, opened 2022-01-17
apidom-ls / apidom - Default to schema element for message payload when empty node


frantuma:
@frantuma commented on [Mon Jan 17 2022](https://github.com/swagger-api/oss-planning/issues/501)

Currently an "empty" payload is not parsed as `schema` element using `refractorPluginReplaceEmptyElement` as done for other empty elements. Goal of this ticket is applying the same mechanism also to payload.

Note. The possible reason for this behavior is `payload` type being `any` therefore admitting also non `schema` types.

```
asyncapi: '2.2.0'
components:
  messages:
    lightMeasured:
      payload:
        
```

frantuma:
slightly related to #1075

char0n:
> AsyncAPI Schema Object = Json Schema Draft 07 + vocabulary

This is only true in specific use-cases. The problem is that AsyncAPI Schema Object replaces every object containing `$ref` with `Reference Object`. JSON Schema Draft 07 does not, and dereferencing and resolution rules are completely different. I've aligned how I do wanna try to approach this problem in https://github.com/swagger-api/apidom/issues/1075#issuecomment-1066602085. Let me know what you think.

frantuma:
@char0n I agree and look forward for https://github.com/swagger-api/apidom/issues/1075#issuecomment-1066602085, as it will address various areas. In scope of my previous comment, applying what mentioned in #1075 would probably address the above also.

PR #1111, opened 2022-01-25
feat: implement plugin to replace empty payload element

frantuma:
Refs #1076


char0n:
Further discussions: https://github.com/swagger-api/apidom/issues/1076#issuecomment-106660284

Issue #1112, opened 2022-01-25
AsyncAPI: Empty nodes support for `Message Object` and `MessageTrait Object` 


char0n:
Continuation of https://github.com/swagger-api/apidom/issues/1076

The goal is to try to integrate support for empty nodes for `Message Object` and `MessageTrait Object` from AsyncAPI 2.2 spec into the existing AsyncAPI empty nodes plugin.

PR #1113, opened 2022-01-25
feat(ns-asyncapi-2): add support for Message.payload empty values

char0n:
Refs #1112
Refs #1076


char0n:
Supersedes https://github.com/swagger-api/apidom/pull/1111

Issue #1582, opened 2022-06-16
apidom-ls: do not validate unknown specs as AsyncAPI 2.x


char0n:
**Current behavior**

I've noticed that `apidom-ls` validates unknown specs as AsyncAPI 2.x. 

**Expected behavior**

`apidom-ls` detects that spec is not AsyncAPI 2.x and will not do any validations, nor other things like autompletion or documentation.

**Fixture that is detected as AsyncAPI 2.x**

```yaml
#%RAML 1.0
title: GitHub API
version: v3
baseUri: https://api.github.com
mediaType:  application/json
securitySchemes:
  oauth_2_0: !include securitySchemes/oauth_2_0.raml
types:
  Gist:  !include types/gist.raml
  Gists: !include types/gists.raml
resourceTypes:
  collection: !include types/collection.raml
traits:
securedBy: [ oauth_2_0 ]
/users:
  type: collection
```

frantuma:
Current code defaults to asyncapi when trying to parse a document with no version set. this is implemented [here](https://github.com/swagger-api/apidom/blob/main/packages/apidom-ls/src/parser-factory.ts#L44) and [here](https://github.com/swagger-api/apidom/blob/main/packages/apidom-ls/src/utils/utils.ts#L734-L770=). `isAsyncDoc` is also atm still used in other parts of code to identify the namespace to load rules for. This is a leftover from POC (considering only oas/asyncapi options), as code should probably provide a `inferNamespace` logic, possibly not only looking at "spec version field" but considering other constructs to infer ns.

When no namespace is inferable / supported, if we don't want to have a default like now, the `parse` logic should be updated and possibly some small adaptation of other code in case this breaks anything.

@char0n please let me know if this is urgent and if you need me to work on it.

char0n:
Last two commits regarding detection of spec and format just landed on master.

frantuma:
#1608 and #1614 implement the changed logic: documents not "detected" as a specific "semantic" namespace (e.g. openapi, asyncapi) are not anymore parsed/validated as asyncapi documents. #1614 further introduces usage of underlying detect capabilities provided by apidom "adapter" packages. 

@char0n @ponelat This has some side effect e.g. in swagger-editor (former "ide"):

1. syntax, docs, completion and validation etc. were previously provided for e.g. a "complete" asyncapi spec, but with no `asyncapi` field set, or set to non-known value. This is not anymore the case, and could potentially be annoying for the user 
[...]

PR #1608, opened 2022-06-17
feat(parser-adapter-json): enhance JSON detection

char0n:
This new detection eliminates some really nonsensical
cases.


char0n:
https://github.com/swagger-api/apidom/issues/1582

PR #1625, opened 2022-06-21
feat(apidom-ls): implement getJsonPointerPosition

frantuma:


Ayaz410808 (approved):

Ayaz410808 (approved):

tim-lai:
@frantuma At a glance, this PR looks ok, and I'll plan on providing a String to the function call, per the test example. However, I'm currently have trouble building apidom monorepo locally. Could you go ahead and merge for now for me to try it out? Thanks!

tim-lai:
@frantuma 

First, with respect to `documentCache`, I'm not sure how to access it from outside of Monaco/React. The `jump-to-path` plugin I'm working on is currently isolated from Monaco Editor, and interaction exists only within SwaggerUI. Open to suggestions on how one might access Monaco workers/memory in such a use case. 

In the meantime, I am trying to directly utilize `getJsonPointerPosition` instead. 

```
const apidomContext = {
    defaultLanguageContent: {
      namespace: 'asyncapi',
    },
  };
  const languageService = getLanguageService(apidomContext);
  try {
    const textDoc = TextDocument.create('foo://spec', 'yaml', 0, editorSpecString);
    const context = {
      format: isJsonDoc(textDoc) ? Format.JSON : Format.YAML,
      baseURI: window.location.href,
    };
    console.log('new textDoc:', textDoc);
    console.log('new context:', context);
    const result = await languageService.getJsonPointerPosition(textDoc, context);
    console.log('test result:', result);
    return { data: 'so far so good' };
  } catch (e) {
    console.log('test catch error', e.message);
    return { error: 'unknown error' };
  }
```

But I'm getting the `api === undefined` case, per screenshot below (`apidom-ls error. no api document`).

apidom-pr-1625

So I'm not sure what else I might need to do to get a properly shaped `ParseResult`, other than maybe trying parsing it first in apidom before sending to `getJsonPointerPosition`?

tim-lai:
@frantuma update: Here's the swagger-editor@next PR, note the code update from earlier comment, although still stuck on the `api === undefined` case : https://github.com/swagger-api/swagger-editor/pull/3244.

tim-lai:
@frantuma another question, would this PR as-is be able to handle something like Object Models in SwaggerUI. I.e. it's not an operation with a path.

frantuma:
@tim-lai merged PR as first step, about your questions above:

> First, with respect to documentCache, I'm not sure how to access it from outside of Monaco/React

I would ask also @char0n opinion here, but a path I might imagine is "initiating" on Swagger UI side by sending the path, but have the monaco/editor/worker side perform the call to language service. Possibly we could have a quick call about that.

> But I'm getting the api === undefined case, per screenshot below (apidom-ls error. no api document).

I believe this is due to the version of the OAS spec (from logs above it looks like 3.0.x). With recent  update of apidom, namespae/version detection is more strict, therefore 3.0 OAS specs do not currently get parsed, please update version to 3.1

> , would this PR as-is be able to handle something like Object Models in SwaggerUI, including a $ref?

Can you elaborate a bit more here? what is that code trying to do?

tim-lai:
> Can you elaborate a bit more here? what is that code trying to do?

Ok, I think we can table this question for now. I'm looking at legacy SwaggerEditor with oas2, then checking `specPath` when Models->Order is clicked. I get the expected array value `['definitions', 'Order']` that matches the definition. I do want to confirm that `apidom-ls` would expect to receive `'/definitions/Order` in this scenario.

There is a snippet removing of trailing '#' in a local $ref, but @ponelat mentioned that the intent of `specPath` was for an actual jsonPointer. Which seems exactly what `apidom-ls` does already.

tim-lai:
> I would ask also @char0n opinion here, but a path I might imagine is "initiating" on Swagger UI side by sending the path, but have the monaco/editor/worker side perform the call to language service. Possibly we could have a quick call about that.

Hmm, perhaps something starting like the following (I added questions in the code comments)?

```
// existing
  async getSemanticTokensLegend() {
    return this._languageService.getSemanticTokensLegend();
  }

// new
  async proposalJumpToPath(document, path) {
    // document should probably be a `uri`,
    // but how would a plugin outside of monaco get the `uri`?
    // separately, should there be an external jumpToPathAdapter (as class)
    // that interfaces with this worker?
    // i.e. src (functional component) -> adapter (class) ->this worker
    return this._languageService.getJsonPointerPosition(document, path);
  }
```

We don't have to worry about SwaggerUI directly, as we wrap this particular component in SwaggerEditor.

tim-lai:
> I believe this is due to the version of the OAS spec (from logs above it looks like 3.0.x). With recent update of apidom, namespae/version detection is more strict, therefore 3.0 OAS specs do not currently get parsed, please update version to 3.1

Hmm, ok. SwaggerUI is unable to render OAS3.1 atm. I'll create a dev mock test button to try it out.

frantuma:
> > I would ask also @char0n opinion here, but a path I might imagine is "initiating" on Swagger UI side by sending the path, but have the monaco/editor/worker side perform the call to language service. Possibly we could have a quick call about that.
> 
> Hmm, perhaps something starting like the following (I added questions in the code comments for @char0n )?
> 
> ```
> // existing
>   async getSemanticTokensLegend() {
>     return this._languageService.getSemanticTokensLegend();
>   }
> 
> // new
>   async proposalJumpToPath(document, path) {
>     // document should probably be a `uri`,
>     // but how would a plugin outside of monaco get the `uri`?
>     // separately, should there be an external jumpToPathAdapter (as class)
>     // that interfaces with this worker?
>     // i.e. src (functional component) -> adapter (class) ->this worker
>     return this._languageService.getJsonPointerPosition(document, path);
>   }
> ```
> 
> We don't have to worry about SwaggerUI directly, as we wrap this particular component in SwaggerEditor.

@tim-lai I am not sure about your proposal above, also as I guess `async getSemanticTokensLegend()` is currently not used. I add here some high level thoughts and a recap for my own clarity, as I am not entirely familiar with current ui/editor/monaco architecture:

1. The use case is about reacting to a UI event  (the `jump to path`) in the editor, where the editor would give focus to the related section.
2. UI is going to extract the json pointer
3. apidom-ls languageService provides `getJsonPointerPosition` which returns a position within the document given a json pointer
4. there can be more `languageService` instances, but to take advantage of its caching mechanism, the same instance should be used across the app
5. the "main" languageService is instantiated within the "ApiDOM Worker", and it cannot be exposed to the code in the UI thread

The idea would be therefore to have UI tell Editor to "jump to path", and editor ask Worker to "do the work"

UI ----- `json-pointer` -----> Editor ----- `postMessage or something` -----> Worker --- `ls.getJsonPointerPosition` -----> languageService

tim-lai:
> recap for my own clarity

This is pretty much accurate to my understanding.

 > proposalJumpToPath

This would be the gist of Worker to `ls.getJsonPointerPosition`. And comments in the snippet are about the preceding step to `postMessage or something` and how to wire it up.

That said, this plugin won't be usable until we get OAS2/OAS3.0 rules then. But could progress a little more to at least POC a solution for `postMessage or something`.

Issue #1695, opened 2022-07-01
apidom-ast: Safari crashes and displays blank page


char0n:
![image](https://user-images.githubusercontent.com/193286/176893829-1032434d-8171-43b2-9ba8-35e71722ac41.png)

Refs SWOS-610

PR #1696, opened 2022-07-01
fix(ast): minic lookbehind regexp to support Safari

char0n:
Refs #1695


packages/apidom-ast/src/yaml/schemas/canonical-format.ts
```diff
@@ -88,10 +88,20 @@ const normalizeLineBreaks = (val: string) => val.replace(/\r\n/g, '\n');
 const preventLineBreakCollapseToSpace = (val: string) => val.replace(/\\\n\s*/g, '');
 
 // collapse line breaks into spaces
-const collapseLineBreakToSpace = (val: string) =>
-  val
-    .replace(/(?<![\n]+)\n([^\n]+)/g, (match: string, p1: string) => ` ${p1.trimLeft()}`)
+const collapseLineBreakToSpace = (val: string) => {
+  /**
+   * Safari doesn't support negative lookbehind, thus we sue mimicking technique:
```
char0n, line 93:
```suggestion
   * Safari doesn't support negative lookbehind, thus we use mimicking technique:
```

PR #1946, opened 2022-08-30
feat(ls): add documentation rules for OpenAPI 3.1.0

tim-lai:
Ref: #1975, #1976, #1977, #1978, #1979, #1980, #1981, #1982, #1983, #1984, #1985

### Description

Render documentation in Monaco for the following elements:
- `externalDocumentation` 
- `license`
- `openapi3_1`
- `operation`
- `paths`
- `pathItem`
- `parameter`
- `responses`
- `server`
- `serverVariable`
- `tag`

refer to comments for additional progressive setup/screenshots/notes.

### Motivation and Context

Build out `apidom-ls` rules for OAS3.1

### How Has This Been Tested?

local. No new tests yet.

### My PR contains...

### My changes...

### Documentation

### Automated tests


tim-lai:
Initial definition (modified fixture):
```yaml
openapi: 3.1.0
info:
  title: deref
  version: 1.0.0
servers:
  - description: local
    url: http://localhost:8082/
paths:
  /a:
    get:
      operationId: aget
      parameters:
        - $ref: '#/components/parameters/userId'
    servers:
    - description: extendedm
      url: http://localhost:8080
    parameters:
    - name: id
    in: path
    description: ID of pet to use
    required: true
    schema:
      type: array
      items:
        type: string  
    style: simple
    post:
      operationId: apost
  /b:
    get:
      operationId: bget
      parameters:
        - $ref: '#/components/parameters/userId'
    post:
      operationId: bpost
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/foo'
components:
  schemas:
    foo:
      type: string
    bar:
      $id: http://localhost:8082/
      type: string
  parameters:
    userId:
      $ref: '#/components/parameters/indirection1'
      description: override
    indirection1:
      $ref: '#/components/parameters/indirection2'
      summary: indirect summary
    indirection2:
      $ref: '#/components/parameters/userIdRef'
      summary: indirect summary
    userIdRef:
      name: userId
      in: query
      description: ID of the user
      required: true
    externalRef:
      $ref: ./ex.json#/externalParameter
      description: another ref
```
apidom-ls-pr-1946-1

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
```diff
@@ -0,0 +1,23 @@
...
+
```
frantuma, line 3: Missing `This object MAY be extended with [Specification Extensions](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#specificationExtensions)` I guess..

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
```diff
@@ -0,0 +1,23 @@
...
+const documentation = [
+  {
+    // OAS 3.1 adds the last sentence vs OAS3.0.3, re: "rules for resolving"
+    target: '$ref',
+    docs: 'Allows for an external definition of this path item. The referenced structure MUST be in the format of a [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#pathItemObject). In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined. See the rules for resolving [Relative References](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#relativeReferencesURI).',
```
frantuma, line 8: link shoul be to 3.1.0 instead of 3.0.3

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  {
+    target: 'summary',
+    docs: 'An optional, string summary, intended to apply to all operations in this path.',
```
frantuma, line 12: Not sure here, as the text matches the one in spec, but maybe there is a comma too much..? `An optional, string summary, intended..` where it could read `An optional string summary, intended..` or something, but comment from non native speaker..

tim-lai: fixed in [15d7f36](https://github.com/swagger-api/apidom/pull/1946/commits/15d7f36994d6758adcae1c8e576d8a55746e65f3)

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
tim-lai: fixed in [15d7f36](https://github.com/swagger-api/apidom/pull/1946/commits/15d7f36994d6758adcae1c8e576d8a55746e65f3)

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
tim-lai: fixed

tim-lai:
@frantuma working on `serverVariables` in https://github.com/swagger-api/apidom/pull/1946/commits/dff8c8e4d8ea14541149b4bb4ee649f3677cbd5f, but I can't get it to display.

If this snippet [here](https://github.com/swagger-api/apidom/blob/dff8c8e4d8ea14541149b4bb4ee649f3677cbd5f/packages/apidom-ls/src/config/openapi/server/documentation.ts#L13) is uncommented, then `server.variables` will render this "target". In addition, syntax highlighting will be present if `servers-variables` is enabled in `symbols.ts` and `config.ts` (currently commented out). Note the plural "s".

apidom-ls-pr-1946-3-serverVariables-as-target



I can verify in apidom-playground, that the element is found and described as `"content": "server-variables"`. Note the singular "server".

I tried different naming conventions in `symbols.ts` and `config.ts`, to no avail. The alternative names are currently commented out.

Here's the minimal spec used for apidom-playground:
```yaml
openapi: 3.1.0
info:
  title: deref
  version: 1.0.0
servers:
  - description: local
    url: http://localhost:8082/
    variables:
      username:
        # note! no enum here means it is an open value
        default: demo
        description: this value is assigned by the service provider, in this example `gigantic-server.com`
      port:
        enum:
          - '8443'
          - '443'
        default: '8443'
      basePath:
        # open meaning there is the opportunity to use special base paths as assigned by the provider, default is `v2`
        default: v2
```

Also the syntax highlight doesn't apply in this screenshot:

apidom-ls-pr-1946-2-serverVariables



Seems like there's a difference between `apidom` parsing as `server-variables` vs `apidom-ls` interpreting as `servers-variables`. Any suggestions on where I can go investigate further and resolve?

frantuma:
@tim-lai 

> If this snippet [here](https://github.com/swagger-api/apidom/blob/dff8c8e4d8ea14541149b4bb4ee649f3677cbd5f/packages/apidom-ls/src/config/openapi/server/documentation.ts#L13) is uncommented, then server.variables will render this "target".

This is correct, as `variables` is a field of `server` and therefore it's picked up. This is however true only if there is no "root docs" defined for `server-variables` element  If you defined (instead) docs for `server-variables`, they will be picked up with preference over the one defined with `target`

> @frantuma working on serverVariables in https://github.com/swagger-api/apidom/commit/dff8c8e4d8ea14541149b4bb4ee649f3677cbd5f, but I can't get it to display.

Not sure here what you mean? display the docs? or? see above

> In addition, syntax highlighting will be present if servers-variables is enabled in symbols.ts and config.ts (currently commented out). Note the plural "s".

I don't think this is correct: `servers-variables` is not a defined "classifier" (either `element` field or member of `classes` of an apidom ns element, see also [here](https://github.com/frantuma/apidom-ls-client/blob/main/README.md#providing-rules)). 

The "classifier" defined for the server-variables element is [server-variables](https://github.com/swagger-api/apidom/blob/a45009fef5855f3b2fba699c9e4faf560c613443/packages/apidom-ns-openapi-3-1/src/elements/nces/ServerVariables.ts#L4). I have seen you added to `tokens.ts` in the branch, and this is correct and sufficient to have it "considered" in syntax highlighting. 

File `symbols.ts` is not related to syntax highlight. I would remove `server-variables` from there, as this is the list of elements considered when `F1->Go to Symbol` is triggered, and I would say server variables would "pollute" the list.

In order to have some visible highlight, obviously it must be configured on the editor side as well

frantuma:
@tim-lai 2 additional not so related comments:

* you added `pathItem-parameters` to `symbols.ts`, again I would remove it because of same reason I mentioned above (also the correct "classifier" would be `path-item-parameters`)
* I would remove the comma from `An optional, string description` in path item description docs, as you did for summary

tim-lai:
@frantuma 

> Not sure here what you mean? display the docs? or? see above

It's not picking up  docs from `src/config/openapi/server-variable/documentation.ts`. 

apidom-ls-pr-1946-3-serverVariables-parent-no-render

I also tried adding the following to `src/config/openapi/config.ts`, without effect
```
serverVariable: serverVariableMeta,
```

tim-lai:
> 2 additional not so related comments:
> 
> * you added `pathItem-parameters` to `symbols.ts`, again I would remove it because of same reason I mentioned above (also the correct "classifier" would be `path-item-parameters`)
> * I would remove the comma from `An optional, string description` in path item description docs, as you did for summary

Resolved. Also removed the above mentioned `server-variables` in `symbols.ts`.

At this moment, only remaining issue is rendering docs from `server-variable` in above comment

frantuma:
@tim-lai 

> It's not picking up docs from src/config/openapi/server-variable/documentation.ts.


Still not clear to me if we are referring to singular (`serverVariable`) or plural (`server-variables`).

A couple of points to recap/clarify:

1. There are still some inconsistencies in naming classifiers. Some are using CamelCase (e.g. `serverVariable`) and some kebab-case (e.g. `server-variables`); this is a known issue which will be fixed.

1. `server-variables`: the `variables` field of Server. As mentioned you can either add docs to server with target `variables` (currently commented out in your branch), or add a "meta" for element `server-variables`; in your branch you have currently [serverVariables](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/config.ts#L36) defined, this should be changed to `'server-variables': serverVariablesMeta,`.

Notice also that you are mixing plural/singular as in [current code](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/config.ts#L36) relates variables to variable (singular) meta. You should instead add a meta directory for `server-variables` and set docs there, and import/use that one in config.

Alternatively, and possibly easier in this case, uncommenting the docs in server for target `variables` would do the work, also removing [line above](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/config.ts#L36).

3. `serverVariable` singular: a variable within the variables. your current code in line 37 should be ok, the docs would appear hovering a variable key

tim-lai:
@frantuma 

> 2\. this should be changed to `'server-variables': serverVariablesMeta,`.

This did the trick! This is my preferred solution, to import the meta docs. PR updated. 

For clarity:
- the meta directory is for `server-variable`, singular. This is consistent with [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#server-variable-object), even though the [Server Object fixed fields](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-4) has a field name for `variables`, plural.
- the element name found by `apidom` is `server-variables`, plural.
- will deal with syntax highlighting for `variables` in SwaggerEditor later.



> 1. There are still some inconsistencies in naming classifiers. Some are using CamelCase (e.g. `serverVariable`) and some kebab-case (e.g. `server-variables`); this is a known issue which will be fixed.

Ok, good to know. Let me know if/when changes are needed. Also, where would I have been able to see these naming classifiers?

frantuma:
@tim-lai not sure if we are there yet

IIANM, your current code would cause the rendering of [this doc](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/server-variable/documentation.ts#L17) when hovering the key `variables`. This is not correct, as the [docs](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-4) for `variables` should be  

```
Map[string, [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject)] 	A map between a variable name and its value. The value is used for substitution in the server's URL template.
```

As mentioned in comment above:

Notice also that you are mixing plural/singular as in [current code](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/config.ts#L36) relates variables to variable (singular) meta. You should instead add a meta directory for server-variables and set docs there, and import/use that one in config.

frantuma:
about "classifiers",  these are either element field or member of classes of an apidom ns element, see also [here](https://github.com/frantuma/apidom-ls-client/blob/main/README.md?rgh-link-date=2022-09-01T10%3A20%3A24Z#providing-rules). 

In local code, the most handy way atm is to search in root path for `this.element = '`, `this.classes.push('` and `primaryClass = '` in `.ts` files

tim-lai:
> IIANM, your current code would cause the rendering of [this doc](https://github.com/swagger-api/apidom/blob/5e701ca221b8a25274016bd72f11da4efc6dfc15/packages/apidom-ls/src/config/openapi/server-variable/documentation.ts#L17) when hovering the key `variables`. This is not correct, as the [docs](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-4) for `variables` should be
> 
> ```
> Map[string, [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject)] 	A map between a variable name and its value. The value is used for substitution in the server's URL template.
> ```

That was my intent, because `Server Variable Object` has its own meta documentation.

If we go back and revisit `pathItem.operation` (paths -> /a -> get), the hover displays the `Operation` meta, and not the table description in [Path Item Object fixed fields](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields-7)

That said, upon further review, `server.variables` is a Map of a string or object, in which case it is more correct to use the table description. There's another case for PathItem.parameters, of an array of Parameter Object OR Reference Object, to which the table description would be more correct to use.

So would this be an acceptable standard?
- if only a single Object type, and meta is available, use the meta. E.g. Operation object
- if array of singe Object type,  and meta is available, use the meta. E.g. Servers object
- if multiple type, use the table description. E.g. Server Variable

fwiw, the Operation Object meta documentation follows this standard, but that's because I haven't gotten to building out the available meta fields for it. :)

Another question, with the above standard, how would a user would visualize a meta data object like `server-variables`, or `tag`, or the Reference Object (which is always paired with a different Object)?

Separately, I'll change the meta directory name to `server-variables`, which would match more closely with `apidom` rather that the literal naming in the specification.

frantuma:
@tim-lai 

 I am having difficulties understanding your point(s):

> That was my intent, because Server Variable Object has its own meta documentation.

A Variable has its own doc, which should be rendered in hover when hovering over the variable key. `variables` has its own docs, as mentioned above, which render in hover when hovering the `variables` field in server.

This is exactly the same as the logic applied to asyncapi, which got its variables and stuff from openapi. Please try hovering the following doc loaded in editor next:

```
asyncapi: 2.4.0
info:
  title: Streetlights Kafka API
  version: 1.0.0
servers:
  test:
    url: test.mykafkacluster.org:8092
    protocol: kafka-secure
    description: Test broker
    variables: 
      foo: 
        description: test
```

> If we go back and revisit pathItem.operation (paths -> /a -> get), the hover displays the Operation meta, and not the table description in [Path Item Object fixed fields](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md?rgh-link-date=2022-09-01T19%3A06%3A59Z#fixed-fields-7)

see above, `operation` is a "single" value representing an operation, comparable to `variable`, not `variables`

> That said, upon further review, server.variables is a Map of a string or object, in which case it is more correct to use the table description. There's another case for PathItem.parameters, of an array of Parameter Object OR Reference Object, to which the table description would be more correct to use.

I am not sure I understand what you mean, please check how done in asyncapi with parameters and references

> So would this be an acceptable standard?

> if only a single Object type, and meta is available, use the meta. E.g. Operation object
> if array of singe Object type, and meta is available, use the meta. E.g. Servers object
> if multiple type, use the table description. E.g. Server Variable

Not sure I understand, puzzled by "multiple type", "table description". I would again point to asyncapi behavior, the "standard" is to use the docs provided by the spec relative to the field. So in your example `Operation` is [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operationObject), `Servers` is [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md?plain=1#L196) and `Server Variable` is [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject)


> fwiw, the Operation Object meta documentation follows this standard, but that's because I haven't gotten to building out the available meta fields for it. :)

Not sure I get this, please look at AsyncAPI operation, this is the same. In your current code it looks ok I guess, with fields docs (with `target`) currently only partially implemented

> Another question, with the above standard, how would a user would visualize a meta data object like server-variables, or tag, or the Reference Object (which is always paired with a different Object)?

Assuming "visualize" means rendering the docs on hover, this is covered by previous points, see asyncapi:

`server-variables`: `Map[string, [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject)] 	A map between a variable name and its value. The value is used for substitution in the server's URL template.`

tag: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#tagObject

$ref: see async, e.g. for [parameter](https://github.com/swagger-api/apidom/blob/7b69f4987d587d873a8b95fa956cad8ded53dd96/packages/apidom-ls/src/config/asyncapi/parameter/documentation.ts#L16)

tim-lai:
@frantuma 

current fixture, up to https://github.com/swagger-api/apidom/pull/1946/commits/e9d8130b116bfc48bdf7d5fc29d94c6ada9e9f5c.

- `paths`
- `pathItem`
- `parameters`
- `servers`
- `server`
- `serverVariable`
- `operation` (now also built out)
- `license`
- `tags`
- `externalDocumentation` 
- `responses`

```yaml
openapi: 3.1.0
info:
  title: deref
  version: 1.0.0
  license:
    name: Apache 2.0
    identifier: Apache-2.0
  contact:
    name: API Support
    url: https://www.example.com/support
    email: support@example.com
servers:
  - description: local
    url: http://localhost:8082/
    variables:
      username:
        # note! no enum here means it is an open value
        default: demo
        description: this value is assigned by the service provider, in this example `gigantic-server.com`
      port:
        enum:
          - '8443'
          - '443'
        default: '8443'
      basePath:
        # open meaning there is the opportunity to use special base paths as assigned by the provider, default is `v2`
        default: v2
tags:
  - name: first tag
    description: basic first tag
  - name: second tag with external
    description: tag description with externalDoc
    externalDocs:
      description: Find out more
      url: http://swagger.io
externalDocs:
  description: externalDocs for main schema
  url: http://localhost:8080
paths:
  /a:
    get:
      operationId: aget
      tags:
        - apple
        - banana
      summary: short explanation
      description: longer explanation
      deprecated: false
      parameters:
        - $ref: '#/components/parameters/userId'
      externalDocs:
        description: externalDocs for operation /get a
        url: http://localhost:8080
      servers:
      - description: extendedm
        url: http://localhost:8080
      - description: secondServ
        url: http://localhost:8081
      variables:
        color:
          default: blue
      responses:
        '200':
          description: a pet to be returned
          content: 
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'
        default:
          description: Unexpected error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorModel'
    post:
      operationId: apost
      parameters:
      - name: id
        in: path
        description: ID of pet to use
        required: true
        schema:
          type: array
          items:
            type: string  
        style: simple
        examples:
        example:
  /b:
    get:
      operationId: bget
      parameters:
        - $ref: '#/components/parameters/userId'
    post:
      operationId: bpost
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/foo'
```

char0n (changes requested):
@tim-lai thanks for starting the work on this. I've finished my deep code review and suggested many changes and enhancements. The goal now should be to stabilize code in this PR, not produce any new stuff and get this PR into state that is mergable, so that rules for OpenAPI 3.0.x production can start in parallel.

Then it would be good if process of producing further documentation for OpenAPI 3.1.0 changes and instead of producing huge PRs, small PRs containing documentation for one/two specification objects would be more appropriate. It would make code reviews more effective and quicker.

packages/apidom-ls/src/config/openapi/operation/documentation.ts
```diff
@@ -0,0 +1,51 @@
...
+    docs: '#### [`string`]\n\nA list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.',
+  },
+  {
+    target: 'summary',
+    docs: 'A short summary of what the operation does.',
+  },
+  {
+    target: 'description',
+    docs: 'A verbose explanation of the operation behavior. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.',
+  },
```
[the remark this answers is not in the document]
tim-lai: @char0n  I can make this change, however, I see that the docs are already picked up from the meta doc:

apidom-ls-pr-1946-comment-operation-externalDocs

packages/apidom-ls/src/config/openapi/path-item/documentation.ts
```diff
@@ -0,0 +1,31 @@
+const docsPathItemObject =
```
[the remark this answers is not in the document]
tim-lai: > We're missing documentation for all the Operation fields like: `get`, `post`, ...etc

These fields will render the docs from the Operation Object meta. 

apidom-ls-pr-1946-comment-pathItem-1


The parent `pathItem` does contain docs for these operations.

apidom-ls-pr-1946-comment-pathItem

packages/apidom-ls/src/config/openapi/tags/documentation.ts
```diff
@@ -0,0 +1,16 @@
+const documentation = [
+  { target: 'name', docs: '**REQUIRED**. The name of the tag.' },
+  {
+    target: 'description',
+    docs: 'A description for the tag. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.',
+  },
+  {
+    target: 'externalDocs',
+    docs: 'Additional external documentation for this tag.',
```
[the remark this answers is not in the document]
tim-lai: Going to remove this field entirely. `External Documentation Object` is already rendered from the meta doc.

apidom-ls-pr-1946-comment-tags-externalDocs

tim-lai:
@char0n Thanks for the detailed review! I've either resolved or commented on all requested changes, please take another look.

Summary:
- add `targetSpecs` anywhere that there is a link to 3.1 in the docs
- moved `tags` and `servers` to `openapi3_1` module
- replaced k/v pair's value with the actual string instead of a `const obj=''`
- commented on `externalDocs`, left as unresolved comment/change
- removed directories no longer needed
- cleanup all comments/placeholders

2 additional questions:
1. CI is reporting commitlint errors, but they don't appear to be from this PR directly. Thoughts? Here's the latest CI run:  https://github.com/swagger-api/apidom/runs/8237567429?check_suite_focus=true
2. I did experiment with adding the documentation in `config/openapi/oas3_1` to `config/common/schema` instead, but the change was not registering. Did not investigate further, as the request was for the `config/openapi/oas3_1` module. But I am curious if there's an obvious reason or something I missed.

char0n:
> @char0n Thanks for the detailed review! I've either resolved or commented on all requested changes, please take another look.
> 
> Summary:
> 
> * add `targetSpecs` anywhere that there is a link to 3.1 in the docs
> * moved `tags` and `servers` to `openapi3_1` module
> * replaced k/v pair's value with the actual string instead of a `const obj=''`
> * commented on `externalDocs`, left as unresolved comment/change
> * removed directories no longer needed
> * cleanup all comments/placeholders

Thanks for changes. Doing second round of CR now.

> 2 additional questions:
> 
> 1. CI is reporting commitlint errors, but they don't appear to be from this PR directly. Thoughts? Here's the latest CI run:  https://github.com/swagger-api/apidom/runs/8237567429?check_suite_focus=true

Nothing to worry about. Only linting of commit messages failed. This PR is going to be squash-merged anyway and new commit msg will be provided.

image


> 2. I did experiment with adding the documentation in `config/openapi/oas3_1` to `config/common/schema` instead, but the change was not registering. Did not investigate further, as the request was for the `config/openapi/oas3_1` module. But I am curious if there's an obvious reason or something I missed.

Let's deal with this  separately after this PR. We can schedule a session where we can talk about rules of configs production.

packages/apidom-ls/src/config/openapi/operation/documentation.ts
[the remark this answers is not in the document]
char0n: Right that's a good point. We have two options here.

### Make it explicit

As per my suggestion. I bet if somebody else will come to this file, he will think that the documentation rule is missing and will try to add it (as me just did).

### Make a comment where field is expected

Instead of defining field we leave the comment to explain how things work so that nobody tries to add the rules later.


---

Personally I would go for - **Make it explicit**. It will be consistent with current AsyncAPI rules. All the fields will be explicit and we don't always need to consult specification document  to find out what is the full enumeration of the fields.  And yes, there will be duplicate documentation strings, but in future we want to eliminate this duplicates by changing documentation rules shape from list to object, which will make ti possible to reference it easily.

packages/apidom-ls/src/config/openapi/openapi3_1/documentation.ts
```diff
@@ -0,0 +1,20 @@
...
+    targetSpecs: [
+      { namespace: 'openapi', version: '3.1.0' },
+      { namespace: 'openapi', version: '3.0.0' },
```
char0n, line 7: Remove the 3.0.0 target spec as the doc rules contains links for 3.1.0
```suggestion
```

packages/apidom-ls/src/config/openapi/openapi3_1/documentation.ts
```diff
@@ -0,0 +1,20 @@
...
+    ],
+  },
+  {
+    target: 'tags',
+    docs: "#### [[Tag Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#tagObject)]\n\nA list of tags used by the document with additional metadata. The order of the tags can be used to reflect on their order by the parsing tools. Not all tags that are used by the [Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operationObject) must be declared. The tags that are not declared MAY be organized randomly or based on the tools' logic. Each tag name in the list MUST be unique.\n",
+    targetSpecs: [
+      { namespace: 'openapi', version: '3.1.0' },
+      { namespace: 'openapi', version: '3.0.0' },
```
char0n, line 15: Remove the 3.0.0 target spec as the doc rules contains links for 3.1.0
```suggestion
```

packages/apidom-ls/src/config/openapi/openapi3_1/documentation.ts
```diff
@@ -0,0 +1,20 @@
+const documentation = [
```
char0n, line 1: Can we enumarate here in comment, what fields are missing?

It is very important not to produce incomplete rules for specification object. If somebody comes after you, the person just does't know if the rules are complete or not and always have to check. The rules of thumb is that when `documentation.ts` file is produced, it is expected to be always complete.

packages/apidom-ls/src/config/openapi/operation/documentation.ts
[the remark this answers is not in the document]
char0n: **Update**: I thought about it again in context of OpenAPI specification after studying what I did in AsyncAPI rules. There are actually rules that governs this.

For example use of `External Documentation Object` within the spec:

image
image
image
image

We can see that the `externalDocs` field description doesn't tell us anything significant. So we can omit the the field and let meta docs show us full documentation for `External Documentation Object`.

If the description would contain some significant information, then what we do is create a link to documentation object and provide a significant description

```js
{
  target: 'externalDocs',
  docs: '[External Documentation Object](link#externalDocumentation)\n\nSignificant description',
}
```

The same applies for example for `Path Item Object` and it's `get`, `post`, `put`, etc... fields. Their description doesn't tell us anything significant, so it's again OK to omit them.

---

To sum it up, for OpenAPI spec it makes sense omit the documentation rules for some fields, but we should put a placeholder comment there instead as suggested by **Make a comment where field is expected**.

char0n:
> commented on externalDocs, left as unresolved comment/change

@tim-lai I've reacted there, have a look at it please.

packages/apidom-ls/src/config/openapi/operation/documentation.ts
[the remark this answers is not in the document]
tim-lai: Done: **Make a comment where field is expected**.

packages/apidom-ls/src/config/openapi/openapi3_1/documentation.ts
```diff
@@ -0,0 +1,20 @@
...
+  {
+    target: 'servers',
```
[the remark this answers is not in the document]
tim-lai: fixed

packages/apidom-ls/src/config/openapi/openapi3_1/documentation.ts
tim-lai: Done:
- Added fields that should render
- Make a comment where field is expected.

All fixed fields referenced [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#fixed-fields) is now listed or commented in `openapi3_1/documentation.ts`

tim-lai:
@char0n 

> After we address rest of this comments, we should be able to go for merge during today/tommorow.

Requested changes should now all be resolved. Please have one more look. Thanks!

Issue #2063, opened 2022-09-23
OpenAPI 3.1.0 language server rules


tim-lai:
The scope of this issue is to provide documentation, completion and linting rules in `apidom-ls` for OpenAPI 3.1.0 specification.

Issue #2062, opened 2022-09-23
OpenAPI 3.1.0 language server completion rules


tim-lai:
The scope of this issue is to provide completion rules in `apidom-ls` for OpenAPI 3.1.0 specification.

## Implementation progress

Only fully implemented specification objects should be checked here.

- [x] [OpenAPI Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#oasObject)
- [x] [Info Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#infoObject)
- [x] [Contact Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#contactObject)
- [x] [License Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#licenseObject)
- [x] [Server Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverObject)
- [x] [Server Variable Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#serverVariableObject)
- [x] [Components](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#componentsObject)
- [x] [Paths Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#pathsObject)
- [x] [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#pathItemObject)
- [x] [Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operationObject)
- [x] [External Documentation Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#externalDocumentationObject)
- [x] [Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameterObject)
- [x] [Request Body Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#requestBodyObject)
- [x] [Media Type Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#mediaTypeObject)
- [x] [Encoding Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#encodingObject)
[...]

PR #2106, opened 2022-10-04
OpenAPI 3.1.0 lint and completion rules for `openapi` object

tim-lai:
### Description

Full lint and completion rules for OpenAPI 3.1.0 `openapi` Object.

- [x] autocompletion for `openapi` object fields
- [x] `info` should be an object, and is required
- [x] `webhooks` should be a  Map of `PathItem` Objects
- [x] `jsonSchemaDialect` should be a `string`
- [x] copy over `openapi3_0` rules but with `openapi3_1` codes. Per comment `This means they don't share any logic.`
- [x] `openapi` contains at least oneOf `paths`, `components`, `webhooks`

also,
- fix error codes for `OPENAPI3_1_INFO` and `OPENAPI3_1_INFO_FIELD_SUMMARY_TYPE` to be consistent with equivalent `OPENAPI_3_1_INFO_XXX` codes
- fix documentation of an unintended partial code snippet formatting
- fix a OpenAPI 3.x documentation comment for `contact` object instead of `context` object

### Motivation and Context

Ref: #2062
Ref: #2063

### How Has This Been Tested?

local testing.

- [x] missing required `info` object should display an error
- [x] `info` object should not be a string (should display error)
- [x] `info` object should be an object
- [x] `info` object should inherit required `title` and `version` fields from OpenAPI 3.0.x rules (should display error if not present)
- [x] `webhooks` should not have an error if it contains a PathItem Object (see fixture)
- [x] `webhooks` should display error if not a Map of PathItem Objects, e.g. user write error as string
- [x] `jsonSchemaDialect` should be a ~~string~~ URI
- [x] `jsonSchemaDialect` should display error if not a `URI`, e.g. user write error as object
- [x] existing `openapi3_0` rules now work for `openapi3_1`. `components--type`, `external-docs--type`, `info--type`, `paths--type`, `security-items-type`, `security--type`, `servers--items`, `servers--type`, `tags--items-type`, `tags--type` 
- [x] `openapi` contains at least oneOf `paths`, `components`, `webhooks`

```yaml
openapi: 3.1.0
info: 
  title: deref
  version: 1.0.0
webhooks:
  # Each webhook needs a name
  newPet:
    # This is a Path Item Object, the only difference is that the request is initiated by the API provider
    post:
      requestBody:
        description: Information about a new pet in the system
jsonSchemaDialect: onetwoString
```

### My PR contains...

### My changes...

### Documentation

### Automated tests


tim-lai:
@char0n fyi, in keeping with producing smaller PRs, I'm thinking of limiting this PR as-is (subject to review feedback), while moving the following item to a separate smaller PR:

> * `openapi` contains at least oneOf `paths`, `components`, `webhooks` --> requires a new linterFunction

packages/apidom-ls/src/config/codes.ts
```diff
@@ -695,8 +695,14 @@ enum ApilintCodes {
 
   OPENAPI3_1_OPENAPI_VALUE_PATTERN_3_1_0 = 7000100,
 
-  OPENAPI3_1_INFO = 7010000,
-  OPENAPI3_1_INFO_FIELD_SUMMARY_TYPE = 7010100,
+  OPENAPI3_1_OPEN_API = 7010000,
```
char0n, line 698: Just a notice here: don't do this in the future - changing the code error numbers. This will shuffle the error numbers that people might be relaying on.

packages/apidom-ls/src/config/openapi/openapi3_1/meta.ts
```diff
@@ -1,8 +1,12 @@
+import completion from './completion';
 import documentation from './documentation';
+import lint from './lint';
 import { FormatMeta } from '../../../apidom-language-types';
 
 const meta: FormatMeta = {
+  completion,
   documentation,
+  lint,
```
char0n, line 9: This is un-important nitpick, that I mention only because of code consistency. The order by convention that we use is:

```js
lint,
completion,
documentation,
```

packages/apidom-ls/src/config/openapi/openapi3_1/lint/jsonSchemaDialect--type.ts
```diff
@@ -0,0 +1,16 @@
+import ApilintCodes from '../../../codes';
```
char0n, line 1: This must be implemented as `json-schema-dialect--format-uri.ts`. 

Notice the fine name convention and the `--format-uri`. The documentation says following about this field: `This MUST be in the form of a URI.`

char0n:
> @char0n fyi, in keeping with producing smaller PRs, I'm thinking of limiting this PR as-is (subject to review feedback), while moving the following item to a separate smaller PR:
> 
> > * `openapi` contains at least oneOf `paths`, `components`, `webhooks` --> requires a new linterFunction

You should finish coverying `openapi3_1` object with rules in this PR, so that we have full picture of rules before meging.

> * `openapi` contains at least oneOf `paths`, `components`, `webhooks` --> requires a new linterFunction

How did you come up with this requirement? I don't see it in the specification document. Can you explain please?

packages/apidom-ls/src/config/openapi/openapi3_1/completion.ts
```diff
@@ -0,0 +1,140 @@
...
+      kind: 'markdown',
+      value:
+        'Map[`string`, [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#pathItemObject) &#124; [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#referenceObject)]\n\\\n\\\nThe incoming webhooks that MAY be received as part of this API and that the API consumer MAY choose to implement. Closely related to the `callbacks` feature, this section describes requests initiated other than by an API call, for example by an out of band registration. The key name is a unique string to refer to each webhook, while the (optionally referenced) Path Item Object describes a request that may be initiated by the API provider and the expected responses. An [example](https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.1/webhook-example.yaml) is available.',
+    },
+  },
+  {
+    label: 'jsonSchemaDialect',
+    insertText: 'jsonSchemaDialect',
+    kind: 14,
+    format: CompletionFormat.OBJECT,
```
char0n, line 129:
```suggestion
    format: CompletionFormat.UNQUOTED,
```

char0n (changes requested):
We're missing bunch of field validations. Please have a look at https://github.com/swagger-api/apidom/tree/09248f68c65129425a8224bcda450399c91ec25d/packages/apidom-ls/src/config/openapi/openapi3_0/lint.

`openapi3_0` and `openapi3_1` are distinct elements. This means they don't share any logic. These two elements are the exception the the rules productions. Other elements only need rules for that divergees between spec versions, but these two needs they own (duplicated) rules.

tim-lai:
> How did you come up with this requirement? I don't see it in the specification document. Can you explain please?

release notes in [3.1.0-rc0](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0)

>An OpenAPI Document now requires at least one of paths, components or webhooks to exist at the top level. While previous versions required paths, now a valid OpenAPI Document can describe only webhooks, or even only reusable components. Thus, an OpenAPI Document no longer necessarily describes an API.

packages/apidom-ls/src/config/openapi/openapi3_1/meta.ts
tim-lai: yea, I went with alphabetical order. will change.

packages/apidom-ls/src/config/codes.ts
tim-lai: noted. Should we create possibly unused codes anyways for OPENAPI3_1, which might help future devs not overlap/overwrite?

```
  OPENAPI3_1_OPEN_API_FIELD_WEBHOOKS_TYPE = 7010100, ---> 7010900 (or 7010800 if below change also made)
  OPENAPI3_1_OPEN_API_FIELD_JSON_SCHEMA_DIALECT_TYPE = 7010200, ---> 7011000 (or 7010900 if below change also made)
```

Is the following intended, where  `OPENAPI3_0_INFO_FIELD_VERSION_TYPE` is given the same code grouping as `OPENAPI3_0_OPEN_API_FIELD` instead of with other `OPENAPI3_0_INFO_FIELD`(s)?

```
OPENAPI3_0_OPEN_API_FIELD_EXTERNAL_DOCS_TYPE = 5010700,
OPENAPI3_0_INFO_FIELD_VERSION_TYPE = 5010800,
OPENAPI3_0_INFO_FIELD_VERSION_REQUIRED,

OPENAPI3_0_INFO = 5020000,
```

vs should be? 

```
OPENAPI3_0_OPEN_API_FIELD_EXTERNAL_DOCS_TYPE = 5010700,

OPENAPI3_0_INFO = 5020000,
OPENAPI3_0_INFO_FIELD_VERSION_TYPE = 5020600,
OPENAPI3_0_INFO_FIELD_VERSION_REQUIRED,

```

If you agree that the codes should be changed, I can create a new ticket for it. And if needed, make a breaking change note in an upcoming release.

packages/apidom-ls/src/config/openapi/openapi3_1/lint/jsonSchemaDialect--type.ts
tim-lai: will fix

packages/apidom-ls/src/config/openapi/openapi3_1/completion.ts
tim-lai: will fix

tim-lai:
> We're missing bunch of field validations.

resolved

char0n:
> > `openapi3_0` and `openapi3_1` are distinct elements. This means they don't share any logic. These two elements are the exception the the rules productions. Other elements only need rules for that divergees between spec versions, but these two needs they own (duplicated) rules.
> 
> Ok thanks for explanation. Will fix

What we can theoretically do to save some code is to import the lint rules from `openapi3_0`. But it's a short-cut that will bite us in future - if we need to support OpenAPI 2.0 or OpenAPI 4.0.0, we would have to copy the rules anyway as the rules would have to be distinguish with `targetSpecs` markers.

char0n:
> > How did you come up with this requirement? I don't see it in the specification document. Can you explain please?
> 
> release notes in [3.1.0-rc0](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0)
> 
> > An OpenAPI Document now requires at least one of paths, components or webhooks to exist at the top level. While previous versions required paths, now a valid OpenAPI Document can describe only webhooks, or even only reusable components. Thus, an OpenAPI Document no longer necessarily describes an API.

Right. The single source of truth for us is the specification - its markdown version. If found this rule specified by the [JSON Schemas](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml#L44). Whenever there is a conflict between the specification and schemas, the specification always takes precedence. This has been already stipulated in multiple discussions like this one: https://github.com/OAI/OpenAPI-Specification/issues/2666

Anyway I found this information within the specification itself here: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#openapi-document, which means we're fine and should implement this lint rule.

In order to implement it you don't necessarily need a new linting function. You can create 3 linting rules with conditional logic as demonstrated here: https://github.com/swagger-api/apidom/blob/1591cfe265c0eba5ac97b3290c56e7242e620881/packages/apidom-ls/src/config/asyncapi/bindings/ibmmq/channel-binding/lint/topic--required.ts#L12

The logic of the condition for each of these rules will be: if other two fields don't exist, this field must be defined. This makes it possible to create overall linting logic as aggregate of 3 separate linting rules.

char0n:
> @char0n fyi, in keeping with producing smaller PRs, I'm thinking of limiting this PR as-is (subject to review feedback), while moving the following item to a separate smaller PR:

up to you. For me the size of this is still comfortable to review as we only deal with `OpenAPI Object` rules alone, so the scope is limited.

tim-lai:
> In order to implement it you don't necessarily need a new linting function. You can create 3 linting rules with conditional logic as demonstrated here:
> 
> https://github.com/swagger-api/apidom/blob/1591cfe265c0eba5ac97b3290c56e7242e620881/packages/apidom-ls/src/config/asyncapi/bindings/ibmmq/channel-binding/lint/topic--required.ts#L12
> 
> The logic of the condition for each of these rules will be: if other two fields don't exist, this field must be defined. This makes it possible to create overall linting logic as aggregate of 3 separate linting rules.

I haven't been able to get this condition example to apply to the openapi object. Latest commit in c599710, with commented out condition. At the moment, each of `paths`, `components`, and `webhooks` are required (without the condition), so I know the rules are recognized. Any suggestions?

In the meantime, I think all other requested changes have been made. Thanks.

char0n:
> I haven't been able to get this condition example to apply to the openapi object. Latest commit in https://github.com/swagger-api/apidom/commit/c59971062a820c24d80fe80ac7d8fe68023a8c9d, with commented out condition. At the moment, each of paths, components, and webhooks are required (without the condition), so I know the rules are recognized. Any suggestions?

Did some experimenting and the easiest solution I could come up with seems to be introduction new rule called `required-fields.ts`:

```ts
import ApilintCodes from '../../../codes';
import { LinterMeta } from '../../../../apidom-language-types';

const requiredFieldsLint: LinterMeta = {
  code: ApilintCodes.OPENAPI3_1_OPEN_API_REQUIRED_FIELDS,
  source: 'apilint',
  message: 'OpenAPI Object must contain one the following fields: paths, components, webhooks',
  severity: 1,
  linterFunction: 'hasRequiredField',
  linterParams: ['paths'],
  marker: 'key',
  conditions: [
    {
      targets: [{ path: 'root' }],
      function: 'missingFields',
      params: [['paths', 'components', 'webhooks']],
    },
  ],
  data: {
    quickFix: [
      {
        message: "add 'paths' section",
        action: 'addChild',
        snippetYaml: 'paths: \n  \n',
        snippetJson: '"paths": {\n  \n  },\n',
      },
    ],
  },
};

export default requiredFieldsLint;
```

packages/apidom-ls/src/config/openapi/openapi3_1/completion.ts
```diff
@@ -0,0 +1,140 @@
...
+      kind: 'markdown',
+      value:
+        'Map[`string`, [Path Item Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#pathItemObject) &#124; [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#referenceObject)]\n\\\n\\\nThe incoming webhooks that MAY be received as part of this API and that the API consumer MAY choose to implement. Closely related to the `callbacks` feature, this section describes requests initiated other than by an API call, for example by an out of band registration. The key name is a unique string to refer to each webhook, while the (optionally referenced) Path Item Object describes a request that may be initiated by the API provider and the expected responses. An [example](https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.1/webhook-example.yaml) is available.',
+    },
+  },
+  {
+    label: 'jsonSchemaDialect',
+    insertText: 'jsonSchemaDialect',
+    kind: 14,
+    format: CompletionFormat.UNQUOTED,
```
char0n, line 129:
```suggestion
    format: CompletionFormat.QUOTED,
```

Before I've suggested to change this from `OBJECT` to `UNQUOTED`. But as this is a string we should change this to `QUOTED`. My mistake.

tim-lai: fixed

tim-lai:
> Did some experimenting and the easiest solution I could come up with seems to be introduction new rule called `required-fields.ts`:

This solution works for me. Implemented in https://github.com/swagger-api/apidom/pull/2106/commits/6d2d7c31632f8ea5ec04b1625807023937f9ba8e

packages/apidom-ls/src/config/openapi/openapi3_1/completion.ts
```diff
@@ -0,0 +1,140 @@
+import {
+  ApidomCompletionItem,
+  CompletionFormat,
+  CompletionType,
+} from '../../../apidom-language-types';
+
+const completion: ApidomCompletionItem[] = [
```
char0n, line 7: just a nitpick here - it's better to define the fields in the same order as they are defined in the spec.

packages/apidom-ls/src/config/openapi/openapi3_1/lint/webhooks--type.ts
```diff
@@ -0,0 +1,18 @@
+// OPENAPI3_1_WEBHOOKS_TYPE
```
char0n, line 1: This should be `webhooks--values-type.ts` - which means that we're linting the type of all the values of the object and not the type of the object itself. See more [here](https://github.com/swagger-api/apidom/tree/975e12e93141897c1ae95ae372037f57a7f83832/packages/apidom-ls/src/config/openapi/components/lint).

packages/apidom-ls/src/config/openapi/openapi3_1/lint/webhooks--type.ts
```diff
@@ -0,0 +1,18 @@
...
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+
+const webhooksLint: LinterMeta = {
+  code: ApilintCodes.OPENAPI3_1_OPEN_API_FIELD_WEBHOOKS_TYPE,
```
char0n, line 6: Note also that if changed to `webhooks--values-type.ts` you need to amend to name of the error code as well.

```
OPENAPI3_1_OPEN_API_FIELD_WEBHOOKS_TYPE -> OPENAPI3_1_OPEN_API_FIELD_WEBHOOKS_VALUES_TYPE
```

packages/apidom-ls/src/config/openapi/openapi3_1/lint/jsonSchemaDialect--format-uri.ts
```diff
@@ -0,0 +1,15 @@
+import ApilintCodes from '../../../codes';
```
char0n, line 1: Name of the file should be `json-schema-dialect--format-uri.ts` (to be consistent with the rest of the codebase).

packages/apidom-ls/src/config/openapi/openapi3_1/lint/jsonSchemaDialect--format-uri.ts
```diff
@@ -0,0 +1,15 @@
...
+import { LinterMeta } from '../../../../apidom-language-types';
+
+const jsonSchemaDialectFormatURILint: LinterMeta = {
+  code: ApilintCodes.OPENAPI3_1_OPEN_API_FIELD_JSON_SCHEMA_FORMAT_URI,
```
char0n, line 5:
```suggestion
  code: ApilintCodes.OPENAPI3_1_OPEN_API_FIELD_JSON_SCHEMA_DIALECT_FORMAT_URI,
```

char0n (dismissed):
We're almost done here. Added couple of requests for changes.

packages/apidom-ls/src/config/openapi/openapi3_1/lint/webhooks--type.ts
char0n, line 1:
```suggestion
```
Unecessary comment here.

tim-lai:
@char0n FYI updated tests in this commit https://github.com/swagger-api/apidom/pull/2106/commits/5e4ad6d233948299ae29ca4dba91e15cecf4a3fb, applies to adding the `openapi3_1` completion rules, though I'm not entirely surely how the expected result would have been pre-determined. Separately, more test fixes incoming for `openapi3_1` linting rules.

@frantuma FYI, this change alleviates the need to be knowledgeable (deepEqual) about specific completion items. We could build out more fixtures and test scenarios if it's really needed, or if there is more that you had originally intended for `CompletionList`. (I'll remove the commented original code before final merge)
```
      // assert.deepEqual(result, input[3] as CompletionList);
      assert(result?.items && result?.items.length > 0);
```

tim-lai:
@char0n Tests updated and are now passing on GA CI. One caveat... the latest commit [eb13b1a](https://github.com/swagger-api/apidom/pull/2106/commits/eb13b1a97c7bd8d9a24d2b5c2c5a6b5316793f03) includes a change that reverts the change from [5e4ad6d](https://github.com/swagger-api/apidom/pull/2106/commits/5e4ad6d233948299ae29ca4dba91e15cecf4a3fb). My local observation is that `npm run test` will now pass as a "bulk" runner when a glob pattern is used, but if one specifies either `openapi-json.ts` or `openapi-yaml.ts` in `mocharc.json` instead of the glob pattern, that specific test suite will fail. The specific test suite case expects an additional two (now commented-out) objects.

tim-lai:
> We're almost done here. Added couple of requests for changes.

done.

packages/apidom-ls/test/openapi-json.ts
```diff
@@ -475,6 +484,40 @@ describe('apidom-ls', function () {
         severity: 1,
         source: 'apilint',
       },
+      // {
```
char0n, line 487: Can you please remove this commented structure? It has been replaced by other up-to-date structures.

packages/apidom-ls/test/openapi-yaml.ts
```diff
@@ -420,6 +621,40 @@ describe('apidom-ls-yaml', function () {
         severity: 1,
         source: 'apilint',
       },
+      // {
```
char0n, line 624: Can you again remove this commented part, as it's no longer used?

packages/apidom-ls/test/openapi-yaml.ts
```diff
@@ -490,10 +725,10 @@ describe('apidom-ls-yaml', function () {
     ];
     assert.deepEqual(result, expected as Diagnostic[]);
 
-    doc = TextDocument.create('foo://bar/specError.yaml', 'yaml', 0, specError);
```
char0n, line 493: I'm not sure I understand this change. Basically this test is no longer running `await languageService.doValidation(doc, validationContext);` at the end. Can you please corroborate why it was removed?

char0n (commented):
Added comments/questions regarding tests. The rest of the code looks LEGIT.

packages/apidom-ls/test/openapi-yaml.ts
tim-lai: There was no followup assertion, so I removed these lines that were leading to it. Note that `openapi-json` doesn't have this.

PR #2155, opened 2022-10-19
fix(ls): finalize rules for OpenAPI License object

char0n:
Refs 20a09d5f778947df0b7f490290381a8c56c1e15c


tim-lai:
@char0n Good catch on this. Thanks for the fixes in this PR.

PR #2196, opened 2022-10-26
feat(ls): provide OpenAPI 3.1.0 completion rules for Responses object

tim-lai:
### Description

* provide OpenAPI 3.1.0 completion rules for Responses object

### Motivation and Context

Ref: #2062 
Ref: #2061 (no update required)

### My PR contains...

### My changes...

### Documentation

### Automated tests


tim-lai:
@char0n I pretty matched the OAS3.0.x version, however, how to best test the response codes? Even without this PR, I'm not getting a suggestion for auto complete in OAS3.0.x definition for `'20...'` (string) or `20...` (number). With partial http code and without closing `:`, I do get the error that `responses` should contain an object.

packages/apidom-ls/src/config/openapi/responses/completion.ts
```diff
@@ -23,211 +23,266 @@ const httpCode3_0CompletionRule = {
...
+  documentation: {
+    kind: 'markdown',
+    value:
+      '[Response Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#responseObject) \\| [Reference Object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#referenceObject)\n\\\n\\\nThe documentation of responses other than the ones declared for specific HTTP response codes. Use this field to cover undeclared responses.',
+  },
+  targetSpecs: [{ namespace: 'openapi', version: '3.1.0' }],
+};
+
 const completion: ApidomCompletionItem[] = [
   {
```
char0n, line 41: You cannot do this, you need to duplicate every rules. After this change things will work as expected

```s
  {
    label: 'default',
    insertText: 'default',
    ...httpCode3_0CompletionRule,
  } as ApidomCompletionItem,
  {
    label: 'default',
    insertText: 'default',
    ...httpCode3_1CompletionRule,
  } as ApidomCompletionItem,

char0n (changes requested):
I've suggested change that will make the things work for you again.

tim-lai:
> I've suggested change that will make the things work for you again.

Changes made. I see completion working when under `paths/path-item/operation/responses`, but not `components/responses`.

char0n:
> > I've suggested change that will make the things work for you again.
> 
> Changes made. I see completion working when under `paths/path-item/operation/responses`, but not `components/responses`.

Here is the type of `Operation Object`.`responses` field

image

Here is the type of `Components Object`.`responses` field

image

Notice the difference. To say it simply - the `Components Object`.`responses` field if of type of generic map/object instead of `Responses Object`. So what you're seeing is correct and this PR can be merged.

char0n (approved):
This is now LEGIT.

Issue #2202, opened 2022-10-27
OpenAPI 3.0.x - completion rules for `$ref`


char0n:
The scope of this issue is to implement completion rules for `$ref` field in all spec objects that supports it.

- [x] Callback Object
- [x] Example Object
- [x] Header Object
- [x] Link Object
- [x] Parameter Object
- [x] Path Item Object
- [x] Request Body Object
- [x] Response Object
- [x] Security Scheme Object

PR #2203, opened 2022-10-27
feat(ls): provide completion rules for $ref field in supported objects

char0n:
Refs #2202

### DIFF
--- a/packages/apidom-ls/src/config/openapi/example/completion.ts
+++ b/packages/apidom-ls/src/config/openapi/example/completion.ts
@@ -6,6 +6,18 @@ import {
 
 const completion: ApidomCompletionItem[] = [
   {
+    label: '$ref',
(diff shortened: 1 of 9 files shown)


packages/apidom-ls/src/config/openapi/example/completion.ts
```diff
@@ -5,6 +5,18 @@ import {
...
+  {
+    label: '$ref',
+    insertText: '\\$ref',
+    kind: 14,
+    format: CompletionFormat.QUOTED,
+    type: CompletionType.PROPERTY,
+    insertTextFormat: 2,
+    documentation: {
+      kind: 'markdown',
+      value: 'A reference to an Example',
```
char0n, line 17:
```suggestion
      value: 'A reference to an Example.',
```

PR #2220, opened 2022-10-30
fix(apidom-ls): remove duplicated rule

frantuma:


char0n (approved):

char0n:
Thanks for catching it @frantuma

PR #2392, opened 2022-12-19
fix(ls): fix OAS reference parameter linting

frantuma:


packages/apidom-ls/src/config/openapi/header/lint/required-fields.ts
```diff
@@ -6,7 +6,7 @@ import { LinterMeta } from '../../../../apidom-language-types';
 const requiredFieldsLint: LinterMeta = {
   code: ApilintCodes.OPENAPI3_0_HEADER_REQUIRED_FIELDS,
   source: 'apilint',
-  message: 'Header Object must contain one the following fields: content, schema',
+  message: 'Header Object must contain one of the following fields: content, schema',
```
char0n, line 9: Header Object can be referenced by Reference Object as well. Shouldn't we add the same condition as in https://github.com/swagger-api/apidom/pull/2392/files#diff-ea2e07d7bd030d4bdfb693da2f82e6126458650b8d2477b077b83b0bc8d71b1fR17?

char0n:
This has been partially addressed in https://github.com/swagger-api/apidom/pull/2545

char0n:
This

PR #2758, opened 2023-05-05
feat(apidom-ls): update exports for JSON Schema validators

frantuma:
missing readme update?

### DIFF
--- a/packages/apidom-ls/package.json
+++ b/packages/apidom-ls/package.json
@@ -20,11 +20,63 @@
   "unpkg": "./dist/apidom-ls.browser.min.js",
   "main": "./cjs/index.cjs",
   "exports": {
-    "types": "./types/dist.d.ts",
-    "import": "./es/index.js",
-    "require": "./cjs/index.cjs"
+    ".": {
+      "types": "./types/dist.d.ts",
+      "import": "./es/index.js",
+      "require": "./cjs/index.cjs"
+    },
+    "./services/validation/providers/json-schema": {
+      "import": "./es/services/validation/providers/json-schema-validation-provider.js",
+      "require": "./cjs/services/validation/providers/json-schema-validation-provider.cjs",
+      "types": "./types/services/validation/providers/json-schema-validation-provider.d.ts"
+    },
+    "./services/validation/providers/asyncapi-20-json-schema": {
+      "import": "./es/services/validation/providers/asyncapi-20-json-schema-validation-provider.js",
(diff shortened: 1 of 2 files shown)


packages/apidom-ls/package.json
```diff
@@ -20,9 +20,61 @@
...
   "exports": {
-    "types": "./types/dist.d.ts",
-    "import": "./es/index.js",
-    "require": "./cjs/index.cjs"
+    ".": {
+      "types": "./types/dist.d.ts",
+      "import": "./es/index.js",
+      "require": "./cjs/index.cjs"
+    },
+    "./services/validation/providers/json-schema-validation-provider": {
```
char0n, line 28: Would be much better to make this export as concise as it can be. 

```suggestion
    "./services/validation/providers/json-schema": {
```

packages/apidom-ls/package.json
```diff
@@ -20,9 +20,61 @@
...
+      "import": "./es/services/validation/providers/json-schema-validation-provider.js",
+      "require": "./cjs/services/validation/providers/json-schema-validation-provider.cjs",
+      "types": "./types/dist.d.ts"
+    },
+    "./services/validation/providers/asyncapi-20-json-schema-validation-provider": {
```
char0n, line 33: Would be much better to make this export as concise as it can be.  This applies to all other exports as well.

```suggestion
    "./services/validation/providers/asyncapi-20-json-schema": {
```

packages/apidom-ls/package.json
```diff
@@ -20,9 +20,61 @@
...
-    "require": "./cjs/index.cjs"
+    ".": {
+      "types": "./types/dist.d.ts",
+      "import": "./es/index.js",
+      "require": "./cjs/index.cjs"
+    },
+    "./services/validation/providers/json-schema-validation-provider": {
+      "import": "./es/services/validation/providers/json-schema-validation-provider.js",
+      "require": "./cjs/services/validation/providers/json-schema-validation-provider.cjs",
+      "types": "./types/dist.d.ts"
```
char0n, line 31: Types must be referenced as individual typing files, not as typing bundle.

```suggestion
      "types": "./types/services/validation/providers/json-schema-validation-provider.d.ts"
```

In order to make this possible this line https://github.com/swagger-api/apidom/blob/1808ea06d75ca5e5c40ddd5719bd01da48fd3758/packages/apidom-ls/package.json#L79 needs to be changed to:

```json
 "types/",
```

char0n (changes requested):
Some changed needed to make mergable. Nice work overally!

Issue #2872, opened 2023-06-19
OpenAPI 3.1.0: `Schema Object`.`example` keyword 


char0n:
In OpenAPI 3.1.0, the `example` field of `Schema Object` is deprecated. The goal of this issue is to add `INFO` or `WARNING` diagnostics serverity linting rule.

Ref SWG-8473

char0n:
Closed by https://github.com/swagger-api/apidom/pull/2909

@kowalczyk-krzysztof nice work!

Issue #2481, opened 2023-01-18
apidom-ls: consolidate linting rules for JSON Schemas


char0n:
The goal of this issue is to consolidate linting rules for JSON Schema in general. This includes rules for Schema Object, that can be found in OpenAPI or AsyncAPI specifications. 

The new system or rules should be flexible, composable and overridable. For example JSON Schema Draft 7 rules, should be composed of Draft 4/5 and Draft 6 rules + overrides. OpenAPI Schema Object rules should be composed of Draft 4/5 rules + overrides.

This architecture will allow us to provide linting for standalone JSON Schemas, not embedded within any other specification.

All the issues found by @frankkilcommins in document called `Specification support within Tools` should be addressed by this issue as well.

Refs https://github.com/swagger-api/apidom/issues/2470

PR #2909, opened 2023-07-05
fix(apidom-ls): add deprecated warning for example property

kowalczyk-krzysztof:
Refs https://github.com/swagger-api/apidom/issues/2872

I'm not sure whether I understood the folder structure properly and if the code is in the right place. I'm also not sure which test file should be changed. I would appreciate some suggestions/help @char0n


frantuma:
@kowalczyk-krzysztof 

`Schema Object` rules are defined in it's [own package](https://github.com/swagger-api/apidom/blob/1b467987f8909c1599feaf36ae4d555ff22268ed/packages/apidom-ls/src/config/common/schema/lint/index.ts), under `common`.

A rule with a warning (possibly better `warning` than `info`?) would look like:

```typescript

import { DiagnosticSeverity } from 'vscode-languageserver-types';

import ApilintCodes from '../../../codes';
import { LinterMeta } from '../../../../apidom-language-types';

const exampleDeprecatedLint: LinterMeta = {
  code: ApilintCodes.SCHEMA_EXAMPLE_DEPRECATED,
  source: 'apilint',
  message: 'property "example" is deprecated, use "examples" instead',
  severity: DiagnosticSeverity.Warning,
  linterFunction: 'missingField',
  linterParams: ['example'],
  marker: 'key',
  markerTarget: 'example',
  targetSpecs: [
    { namespace: 'openapi', version: '3.1.0' },
  ],
  data: {
    quickFix: [
      {
        message: 'remove example',
        action: 'removeChild',
        functionParams: ['example'],
        target: 'parent',
      },
      {
        message: "add 'examples' field",
        action: 'addChild',
        snippetYaml: 'examples: \n  ',
        snippetJson: '"examples": ,\n    ',
      },
    ],
  },
};

```

The sample code includes `targetSpecs` to specify to apply the rules only for OAS 3.1 (as schema is currently "shared" also with OAS 3.0 and AsyncAPI.
The sample also includes some possible quick fixes in case.

The current standard way to add rules to `apidom-ls` "core" is to organize them within the mentioned structure, just adding that it is also possible to add rules with a `given` field external to the dir structure, see e.g. [this test](https://github.com/swagger-api/apidom/blob/fb69f9dd2d35a737bb1100af30d6e52bf05b57ba/packages/apidom-ls/test/validate.ts#L2856). This can be useful to add custom rules in a client env or to do some testing.

In terms of tests, current status is "sub-optimal" for `apidom-ls` package, tests are not well designed/organized.

That said, an option is adding a test e.g. [here](https://github.com/swagger-api/apidom/blob/fb69f9dd2d35a737bb1100af30d6e52bf05b57ba/packages/apidom-ls/test/validate.ts#L2856) (currently it compare stringified diagnostics result with hardcoded string.



for future ref, slightly related to https://github.com/swagger-api/apidom/issues/2481

kowalczyk-krzysztof:
@frantuma Thank you. I pushed the recommended changes and added test.

packages/apidom-ls/src/config/openapi/components/lint/example--deprecated.ts
```diff
@@ -0,0 +1,34 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
```
frantuma, line 1: This should be moved to [src/config/common/schema/lint](https://github.com/swagger-api/apidom/tree/aa317d408ce3687ff3dfa9bb1a14d7c6486891bf/packages/apidom-ls/src/config/common/schema/lint)

kowalczyk-krzysztof: I thought I did that already. My bad. Fixed.

frantuma (approved):

kowalczyk-krzysztof:
The push after approval is just squashing two commits into one, as for some reason `Lint commit` would respond with:
```
header must not be longer than 69 characters, current length is 70 [header-max-length]
```
Even though the actual length was 40 characters.
Same issue happened when I created a merge commit to merge main into my branch - which I fixed by rebasing instead of merging.

PR #2951, opened 2023-07-14
revert: use babel plugin to capture error causes

char0n:
Refs 19171aa
Refs https://github.com/niksy/babel-plugin-native-error-extend/issues/4

### DIFF
--- a/packages/apidom-reference/src/util/errors/CausedError.ts
+++ b/packages/apidom-reference/src/util/errors/CausedError.ts
@@ -1,0 +1,18 @@
+import { hasIn } from 'ramda';
+import { isPlainObject } from 'ramda-adjunct';
+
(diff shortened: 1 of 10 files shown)


packages/apidom-reference/src/util/errors/CausedError.ts
```diff
@@ -0,0 +1,18 @@
+import { hasIn } from 'ramda';
+import { isPlainObject } from 'ramda-adjunct';
+
+class CausedError extends Error {
```
char0n, line 4:
```suggestion
/**
 * This needs to stay here until our minimum supported version of Node.js is >= 16.9.0.
 * Node.js is >= 16.9.0 supports error causes natively.
 */
class CausedError extends Error {
```

PR #2956, opened 2023-07-16
ci(build): fix branch in lint commit messages

frantuma:
@char0n 

lint commit messages job seems to target the wrong commits, including commits not belonging to the branch, see e.g. [this run](https://github.com/swagger-api/apidom/actions/runs/5565788003/jobs/10166426126?pr=2954)

This seems to be related to the check out (by default) of the merge branch instead of the head one, which does mess up somehow commit history, so that  `HEAD~${{ github.event.pull_request.commits }}` for some reason points to a previous non-branch commit. Still kind of mysterious as adding a `git log` to the `HEAD~${{ github.event.pull_request.commits }}`  does point to yet another commit in history..

There might be a better solution, however this PR addresses it by having a separate job `lint-commit-message` do the work, checking out head branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Lint commit message
         if: github.ref != 'refs/heads/main' && github.actor != 'dependabot[bot]'
         run: |
-          npx commitlint --from HEAD~${{ github.event.pull_request.commits }} --to HEAD --verbose
+          npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
 
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,13 +13,32 @@ env:
   CPU_CORES: 2
 
 jobs:
-  lint:
+  lint-commit-messages:
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0 # necessary for linting commit messages
+          ref: ${{ github.event.pull_request.head.sha }} # necessary to check out pull request HEAD commit instead of merge commit
+      - name: Use Node.js 20
+        uses: actions/setup-node@v3
+        with:
+          node-version: 20
+          cache: 'npm'
+          cache-dependency-path: 'package-lock.json'
+      - name: Install dependencies
+        run: npm ci
+      - name: Lint commit message
+        if: github.ref != 'refs/heads/main' && github.actor != 'dependabot[bot]'
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -28,11 +29,26 @@ jobs:
           cache-dependency-path: 'package-lock.json'
       - name: Install dependencies
         run: npm ci
-      - name: Lint code
-        run: npm run lint
       - name: Lint commit message
         if: github.ref != 'refs/heads/main' && github.actor != 'dependabot[bot]'
-        run: npx commitlint --from HEAD~${{ github.event.pull_request.commits }} --to HEAD
+        run: |
+          npx commitlint --from HEAD~${{ github.event.pull_request.commits }} --to HEAD --verbose
```
char0n, line 35: image


```suggestion
          npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
```

Source: https://stackoverflow.com/questions/67307001/how-to-run-commitlint-in-github-workflow-on-every-commit-of-a-push

char0n (approved):

char0n:
Example of success PR: https://github.com/swagger-api/apidom/actions/runs/5574091904/jobs/10182242045?pr=2957

char0n:
Example of failing PR: https://github.com/swagger-api/apidom/actions/runs/5574142702/jobs/10182361222?pr=2958

.github/workflows/build.yml
frantuma: @char0n 

`only commits in scope of the PR will be checked`: IMO this is desired IIANMS; e.g. in [mentioned run](https://github.com/swagger-api/apidom/actions/runs/5565788003/jobs/10166426126?pr=2954) the issue is that a [previous commit](https://github.com/swagger-api/apidom/commit/c46ae6a3b7eb511e38f7576bfc0bff1a8bf4e9ae) NOT in PR scope fails the lint; this commit should have failed the lint but didn't for some reason, but further PRs shouldn't fail because of that..

`We need the base-SHA as well as the head-SHA`: this is true but again IIANM in my experiments this didn't work as possibly the `github.event.pull_request.base.sha` doesn't point to the correct commit, at least in some scenarios. See e.g. [this discussion](https://github.com/orgs/community/discussions/39880) and related [ticket](https://github.com/actions/runner/issues/1689)

I can try experimenting a bit more with the above variables, but again possibly these tests won't cover all scenarios (other commits to base branch, etc), while the proposed solution based on `github.event.pull_request.commits` and PR head commit seems to work in all scenarios

frantuma:
just for ref, in case of further issues see [this comment](https://github.com/swagger-api/apidom/pull/2956#discussion_r1265123819)

PR #2980, opened 2023-07-24
fix(ls): temporary quick fix false positive for schema reference

frantuma:
@char0n  

Refs https://github.com/swagger-api/swagger-editor/issues/3722 
Refs https://github.com/swagger-api/swagger-editor/issues/4026

This is a temporary quick fix to avoid false positives in OAS 3.0.x `anyOf`, `oneOf` and `allOf` constructs, until the "real" underlying issue in schema parsing in OAS 3.0 spec is resolved (e.g. adding `referenced-element` meta or other solution)

This fix would cause a reference to a non schema not to be considered as error for OAS 3.0, but that missing negative is IMO less annoying for users as the current false positive


frantuma:
replaced by #2987 which fixes https://github.com/swagger-api/swagger-editor/issues/3722 and https://github.com/swagger-api/swagger-editor/issues/4026

PR #2982, opened 2023-07-24
fix(ls): process schema 'null' / 'nullable' depending on spec version

frantuma:
Differentiates behavior for type `null` validation/completion depending on spec/version, also add validation rule for `nullable` adding hint for all asyncapi and openapi spec versions except openapi 3.0.x. AsyncAPI also included as users can typically come from openapi 3.0 when designing an AsyncAPI spec


packages/apidom-ls/src/config/common/schema/lint/index.ts
```diff
@@ -56,6 +57,7 @@ import thenNonIfLint from './then--non-if';
 import thenTypeLint from './then--type';
 import titleTypeLint from './title--type';
 import typeTypeLint from './type--type';
+import typeTypeLintOpenApi30 from './type--type-openapi-3-0';
```
char0n, line 60: Just consistent naming with the rest of config
```suggestion
import typeTypeOpenAPI3_0Lint from './type--type-openapi-3-0';
```

packages/apidom-ls/src/config/common/schema/lint/nullable--not-recommended.ts
```diff
@@ -0,0 +1,37 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+
+const nullableNotRecommendedLint: LinterMeta = {
+  code: ApilintCodes.SCHEMA_EXAMPLE_DEPRECATED,
```
char0n, line 7: new code needs to be created and assigned here

packages/apidom-ls/src/config/common/schema/lint/nullable--not-recommended.ts
```diff
@@ -0,0 +1,37 @@
...
+  source: 'apilint',
+  message: 'nullable has no special meaning, if not set on purpose use "type=null" instead',
```
char0n, line 9: value of type must always be string even in case of the null type. This will avoid some confusion that this message might trigger
```suggestion
  message: 'nullable has no special meaning, if not set on purpose use `type="null"` instead',
```

packages/apidom-ls/src/config/common/schema/lint/index.ts
```diff
@@ -120,6 +123,7 @@ const schemaLints = [
   thenTypeLint,
   titleTypeLint,
   typeTypeLint,
+  typeTypeLintOpenApi30,
```
char0n, line 126:
```suggestion
  typeTypeOpenAPI3_0Lint,
```

packages/apidom-ls/src/config/common/schema/lint/type--type-openapi-3-0.ts
```diff
@@ -0,0 +1,57 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+
+const typeTypeLintOpenApi30: LinterMeta = {
```
char0n, line 6:
```suggestion
const typeTypeOpenAPI3_0Lint: LinterMeta = {
```

packages/apidom-ls/src/config/common/schema/lint/index.ts
frantuma: @char0n 
this doesn't seem to be liked by linter: 

 Variable name `typeTypeOpenAPI3_0Lint` must match one of the following formats: camelCase, PascalCase, UPPER_CASE

char0n: Yep, we disable it inline as in https://github.com/swagger-api/apidom/blob/c169eb3c5b00d2b2c6ecad34acb3304da7b3043d/packages/apidom-ls/src/config/openapi/header/lint/content--allowed-fields-3-0.ts#L6

We use this convention for element names containing version numbers as well.

PR #3084, opened 2023-08-28
feat(ls): app option for apidom-reference based ref validation

frantuma:
This PR adds the option to use [apidom-reference](https://github.com/swagger-api/apidom/tree/main/packages/apidom-reference) to perform validation of references (local and external), replacing current local only "legacy" validation

The option is not activated by default as:

1. Error messages are still too technical/low level and refer to the reference node where the error occurs with no further info to identify it, e.g. an error is some external file not part of the root document
2. It may have some performance implications

When the above will be addressed/fixed we will probably get rid of legacy validation and switch fully to `apidom-reference` based

Setting `ValidationContext.referenceValidationMode` controls reference validation:

`referenceValidationMode = ReferenceValidationMode.LEGACY` : Legacy local only validation (default)
`referenceValidationMode = ReferenceValidationMode.APIDOM_INDIRECT` : `apidom-reference` validation of local references
`referenceValidationMode = ReferenceValidationMode.APIDOM_INDIRECT_EXTERNAL` : `apidom-reference` validation of local 
 and external references

Setting `ValidationContext.referenceValidationSequentialProcessing = true` causes reference validation to be processed sequentially instead of concurrently. In some scenarios (e.g. many non-distinct external refs) this might be more efficient.


frantuma:
For future ref, alternatives for `validateReferences`:

1. [concurrent with no cache of refSet](https://gist.github.com/frantuma/7452cca74519ebe9a66246623b07c58c)
2. [sync with cache of refSet](https://gist.github.com/frantuma/b34092c6222074e16eea5dfc69062d5e)

Issue #3198, opened 2023-10-02
Use `toValue()` function instead of `.toValue()` method


char0n:
`toValue` function from `@swagger-api/apidom-core` can handle cycles, `.toValue()` cannot. We shouldn't use `.toValue()` any more, as it's not safe. The goal of this issue is to migrate to `toValue` function all the packages within monorepo.

char0n:
All references of `.toValue()` method replaced for `toValue` function.

PR #3207, opened 2023-10-02
refactor(reference): replace toValue method with function

char0n:
Refs #3198


packages/apidom-ls/test/validate-valid.ts
```diff
@@ -22,6 +22,8 @@ describe('apidom-ls-validate', function () {
 
   // eslint-disable-next-line mocha/no-skipped-tests
   it('test validation of supposedly valid specs', async function () {
+    this.timeout(3000);
```
char0n, line 25:
```suggestion
    this.timeout(10000);
```

PR #3271, opened 2023-10-15
fix: sanitize() double encode special characters

### LINKED ISSUE #3270: Double encoding when sanitized URL with encoded symbols like %2F (/), %3B (;) etc. provided to sanitize() function
### Q&A (please complete the following information)
 - ApiDOM version: 0.77.0
### Describe the bug you're encountering
The bug is follow up of the by reported for SwaggerUI https://github.com/swagger-api/swagger-ui/issues/9304 where it causes the practical problem.

Calling the `sanitize()` function in  https://github.com/swagger-api/apidom/blob/main/packages/apidom-reference/src/util/url.ts#L248 cause double encoding when URL already contains encoded 'special' symbols: `; / ? : @ & = + $ , #`

### To reproduce...

1. Call `sanitize()` function with URL that contains encoded symbols %2F , %3B etc. Similar to test case https://github.com/swagger-api/apidom/blob/main/packages/apidom-reference/test/util/url.ts#L294 . For example `sanitize('https://example.com/path%20with%20spaces%2Fslashes%3Bsemicolons/?including=in%3Fparameters')`
2. The expected output is the same URL: `'https://example.com/path%20with%20spaces%2Fslashes%3Bsemicolons/?including=in%3Fparameters`. 
The real output is this: `https://example.com/path%20with%20spaces%252Fslashes%253Bsemicolons/?including=in%253Fparameters` . 
The `%` of the `%2F` and `%3B` are encoded. It should not happen.

### Expected behavior

The encoded URL passed to `sanitize()` should remain the same for all possible combinations of encoded symbols

### Additional context or thoughts
The reason for the issue is that `decodeURI()` and `encodeURI()` leave 'special' symbols `; / ? : @ & = + $ , #` encoding unchanged. It is mentioned in the documentation at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI#description and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#description .

stepashka69:
### Description
Use `URL().toString()` to construct sanitised URI (including support of IP v6 URIs).
Fall back to the existent functionality in cases when URL() can not parse the URI.

### Motivation and Context
The problem is described in #3270 .
Existent sanitize implementation doesn't return correct result when the URI contains encoded 'special' characters such as `; / ? : @ & = + $ , #`.
Closes #3270 

### How Has This Been Tested?
Tested with existent unit tests. 
Extended test for the case with special characters and added test case for IP v6.

### My PR contains...

- [x] Bug fixes (non-breaking change which fixes an issue)

### My changes...

### Documentation

### Automated tests
- [x] My changes can and should be tested by unit and/or integration tests.
- [x] If yes to above: I have added tests to cover my changes.
- [x] If yes to above: I have taken care to cover edge cases in my tests.


char0n:
@stepashka69 looks good. There is additional [unsanitize](https://github.com/swagger-api/apidom/blob/b7696c6e41c2f9ec5a0cd70fe877ff5f3fed3b95/packages/apidom-reference/src/util/url.ts#L267) function, which needs to be align with the `sanitized` function now.

char0n (approved):
Nicely done!

PR #3361, opened 2023-11-06
fix(reference): fix handling cycles in all dereference strategies

char0n:
Refs https://github.com/swagger-api/swagger-ui/issues/9337


char0n:
apidom-ls is using frozen ApiDOM which causes dereferencing to fail, as the dereferencing tries to manage identity of elements by assiging identity if it doesn't exist. This will fail on frozen ApiDOM. In order to bypass it, I've created a formalized immutable identity management that will be used in this PR: https://github.com/swagger-api/apidom/pull/3366

Issue #3392, opened 2023-11-14
Workflows Specification 1.0.0 - create namespace


frankkilcommins:
Scope of this issue is to implement Workflows 1.0.0 namespace package.

**TODO**:

- [x] [create new empty ApiDOM package called `apidom-ns-workflows-1`](https://github.com/swagger-api/apidom/pull/2528)
- [x] implement refractor layer ([current progress here](https://github.com/swagger-api/apidom/tree/main/packages/apidom-ns-workflows-1#implementation-progress))

PR #3429, opened 2023-11-21
feat(ns-workflows-1): Add Failure Action Obj

frankkilcommins:
Refs: #3392

Add Failure Action Object to apidom-ns-workflows-1


char0n (approved):
Looks very good!

PR #3446, opened 2023-11-23
feat(ns-workflows-1): Add Reference Object

frankkilcommins:
Refs: #3392

Add Reference Object to apidom-ns-workflows-1


packages/apidom-ns-workflows-1/src/elements/Reference.ts
```diff
@@ -0,0 +1,27 @@
+import { StringElement, ObjectElement, Attributes, Meta } from '@swagger-api/apidom-core';
+
+class Reference extends ObjectElement {
+  constructor(content?: Record<string, unknown>, meta?: Meta, attributes?: Attributes) {
+    super(content, meta, attributes);
+    this.element = 'reference';
+    this.classes.push('workflowsSpec-reference');
```
char0n, line 7: Dash case convention for meta classes. Wouldn't `workflows-reference` suffice?
```suggestion
    this.classes.push('workflows-spec-reference');
```

char0n (approved):
LGTM.

Issue #3478, opened 2023-11-30
OpenAPI 2.0: scope OpenAPI 3.x.y linting rules explicitly to supported versions


char0n:
All the OpenAPI 3.x.y needs to be scopes explicitly (using `targetSpecs`), unless they are really specific to OpenAPI unique elements. This will allow to start producing proper OpenAPI 2.0 linting rules.

This is continuation of https://github.com/swagger-api/apidom/issues/3103

char0n:
OpenAPI 3.x.y config is now scoped explicitly.

PR #3505, opened 2023-12-05
refactor(ls): scope OpenAPI 3.x.y rules explicitly (batch 4)

char0n:
Inclues:

- Request Body Object
- Response Object
- Responses Object
- Security Requirement Object
- Security Scheme Object
- Server Object
- Server Variable Object
- Tag Object
- XML Object

Refs #3478


char0n (commented):
Self-review performed

Issue #3479, opened 2023-11-30
OpenAPI 2.0: add support for symbols


char0n:
Some symbol support has already been added in https://github.com/swagger-api/apidom/pull/3470.

It needs to be further reviewed and additional symbols needs to be added.

This is continuation of https://github.com/swagger-api/apidom/issues/3103

char0n:
Closed by https://github.com/swagger-api/apidom/pull/3507

PR #3507, opened 2023-12-05
feat(ls): add missing OpenAPI 2.0 symbols

char0n:
Refs #3479


packages/apidom-ls/src/config/symbols.ts
```diff
@@ -1,42 +1,8 @@
...
-  'responsesDefinitions',
-  'parametersDefinitions',
-  'definitions',
-];
+import adsSymbols from './ads/symbols';
+import asyncapiSymbols from './asyncapi/symbols';
+import openapiSymbols from './openapi/symbols';
+
+// creating list of unique symbols
+const symbols = Array.from(new Set([...adsSymbols, ...asyncapiSymbols, openapiSymbols]));
```
char0n, line 6:
```suggestion
const symbols = Array.from(new Set([...adsSymbols, ...asyncapiSymbols, ...openapiSymbols]));
```

PR #3514, opened 2023-12-08
feat(apidom-ls): validate every param defined

kowalczyk-krzysztof:
Refs SWG-9290
Refs https://github.com/swagger-api/apidom/issues/3517

I know this is not pretty but I wanted to get out something that kinda works for a code review.

Issues:
- a lot of iterations,
- I'm not sure if I'm handling working with the `Element` object correctly,
- I think this will return only one message for the whole definition
- types

I could use some feedback @char0n @frantuma


char0n:
Hi @kowalczyk-krzysztof,

> I know this is not pretty but I wanted to get out something that kinda works for a code review.

Yeah that's a healthy approach.

> a lot of iterations

This is one of the more complex rules to implement and required many moving parts.

---

You've choose to implement really complex rule. I've created an [associated GitHub issue](https://github.com/swagger-api/apidom/issues/3517). where I'll document steps necessary to achieve your goal.

char0n:
Superseded by https://github.com/swagger-api/apidom/pull/3532

PR #3530, opened 2023-12-13
feat(parser-adapter-workflows-1): Add JSON and YAML parser adapters

frankkilcommins:
Add:
- apidom-parser-adapter-workflows-json-1
- apidom-parser-adapter-workflows-yaml-1

PR #3531, opened 2023-12-13
feat(parser-adapter-workflows-json-1): Add adapter

frankkilcommins:
Add JSON parser adapter for the Workflows Specification


char0n:
As discussed we'll use https://github.com/swagger-api/apidom/pull/3530 as this one doesn't pass TypeScript type checker.

Issue #3517, opened 2023-12-11
OpenAPI 2.0/3.x.y - validation of Path Templating


char0n:
We are currently not able to validate and resolve `Path Templating`. I have created low level parser, validator and resolver library for OpenAPI Path Templating here: https://github.com/char0n/openapi-path-templating

`openapi-path-templating` is applicable for: OpenAPI **2.0** and **3.x.y**.

We want to use it for every case related to `Path Templating`. 

> [!WARNING]  
> `openapi-path-templating` library doesn't handle computing appropriate `Parameter Object`s to be resolved against the template expression.

---

I actually see two different linting rules here and I do understand and approach them differently (correct me if I'm wrong).

### Rule 1

We want a linting rule that validates if the template expressions defined within the URL are all present in the appropriate `Parameter Objects`. 

### Rule 2

We want a linting rule that validates that user did not forget to define template expressions for all the appropriate `Parameter Objects`.

For the purpose of this implementation I will assume we're talking about **Rule 1**. 

---

There are multiple layers that we need to implement to achieve the final goal described in **Rule 1**.

### Low level `Path Templating` handling

I've already had a code for this on my local dev computer. I've consolidated the code and pushed with (along with the documentation) to https://github.com/char0n/openapi-path-templating. The library is not yet released, I will release it probably during today. The library handles all the low level complexities and corner cases of parsing, validating and resolving the Path Templating.

The library can distinguish if the URL contains Path Templating or not.

![image](https://github.com/swagger-api/apidom/assets/193286/43fb29a4-8d2d-467b-a250-274b41b91e6a)
> Above is an excerpt from OpenAPI 3.x.y spec, which tells us how to handle escaping of "generic syntax" characters

### High level `Path Templating` handling

![image](https://github.com/swagger-api/apidom/assets/193286/a9a50f21-6fc0-4d64-956c-8979b29496ab)
> Above OpenAPI 3.x.y excerpt tells us how to handle `Path Templating`

#### Collecting `Parameter Objects`

So, what we need to do it to collect all `Parameter Objects` from `Path Item Object` directly mapped to URL with `Path Templating` and  all `Parameter Objects` from every `Operation Object` mapped from `Path Item Object` under the `HTTP verb`. (I'll help you with querying those out from ApiDOM later). 

When we have all the `Parameter Objects` we have to apply following algorithms defined by OpenAPI:

**Applies to both `PathItem.paramters` and `Operation.parameters`**

[...]

PR #3538, opened 2023-12-14
feat(apidom-ls): path template match parameter object lint

kowalczyk-krzysztof:
Refs #3517


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1009,37 @@ export const standardLinterfunctions: FunctionItem[] = [
     functionName: 'apilintOpenAPIPathTemplateValid',
     function: (element: Element) => {
       if (isStringElement(element)) {
-        return true;
+        const parseResult = parse(toValue(element));
+        const parts: string[] = [];
+        parseResult.ast.translate(parts);
+
+        const templateExpressions = parts
```
char0n, line 1016: We can safely drop lines 1016-1022. This is already guaranteed by condition to this rule: https://github.com/swagger-api/apidom/blob/b0b03a2d2549a68701248909d7b73ec5a1d51ae5/packages/apidom-ls/src/config/openapi/path-template/lint/value--valid.ts#L15

kowalczyk-krzysztof: You mean this part?
```typescript
        if (templateExpressions.length === 0) {
          return true;
        }
```
I still need the template expressions list so this:
```typescript
const templateExpressions = parts
          .filter((part) => part[0] === 'template-expression')
          .map((part) => part[1].slice(1, -1));
``` 
has to stay

char0n: You're right. We can drop just part of the code.

kowalczyk-krzysztof: Sure, removed.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1009,33 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          .filter((part) => part[0] === 'template-expression')
+          .map((part) => part[1].slice(1, -1));
```
char0n, line 1018: Would it be possible to be more semantic? By `more semantic` I mean not getting outside of the real of ApiDOM (by converting to  POJO). Performance will greatly suffer as Path Items and Operations are the biggest objects. Ideally we want to stay inside ApIDOM all the time and only do minimal necessary  ApiDOM -> POJO conversions.

I've drafted the code that is semantic does minimal conversions and works on top of ApiDOM:

```js
function: (element: Element) => {
      if (isStringElement(element)) {
        const parameterElements: any[] = [];

        // we're treating PathItemElement polymorphically as ObjectElement as this function has to support all OpenAPI versions
        const pathItemElement = (element.parent as MemberElement).value as ObjectElement;
        if (isArrayElement(pathItemElement.get('parameters'))) {
          parameterElements.push([...pathItemElement.get('parameters')]);
        }

        // special case for handling empty PathItemElement
        if (pathItemElement.isEmpty) {
          return true;
        }

        // we're asserting OperationElement polymorphically as this function has to support all OpenAPI versions
        const operationElements = Array.from(
          pathItemElement.filter((elm: Element) => elm.element === 'operation').map((elm) => elm),
        );
        // collect all ParameterElements from all OperationElements
        operationElements.reduce((params, operationElement) => {
          if (isArrayElement(operationElement.get('parameters'))) {
            params.push(Array.from(operationElement.get('parameters')));
          }
          return params;
        }, parameterElements);

        // filter only appropriate ParameterElements
        const allowedLocation = ['path', 'query'];
        const applicableParameterElements = parameterElements.filter((parameterElement) =>
          allowedLocation.includes(toValue(parameterElement.in)),
        );

        // prepare for path template resolution
        const pathTemplateResolveParams = applicableParameterElements.reduce(
          (params, parameterElement) => {
            params[toValue(parameterElement.name)] = 'placeholder'; // eslint-disable-line no-param-reassign
            return params;
          },
          {},
        );
        const pathTemplate = toValue(element);
        const resolvedPathTemplate = resolve(pathTemplate, pathTemplateResolveParams);

        return test(resolvedPathTemplate, { strict: true });
      }
      return true;
    },
```

Have a look at it. It also handles the case of empty `Path Item Object`. The code also abstracts us from `openapi-path-templating` AST and instead we use it's functions.

char0n (commented):
Left one bigger comment with the draft of possible semantic implementation without doing ApiDOM -> POJO conversions.

packages/apidom-ls/src/services/validation/linter-functions.ts
char0n: There is still lot of stuff missing. I can think of `parameterElements.push`. Before we push anything to `parameterElements` we must make sure that what does inside is `ObjectElement` and it's `element` property equals `parameter`.

char0n: Another thing is that I'm not sure this will handle cases where `Reference Object` is used instead of `Parameter Object`. We will know as long as we have a test case for it. It might be possible that @frantuma already did changes so that linting is performed on dereferenced structure.

kowalczyk-krzysztof: Right. I agree it's better to stay inside ApiDOM. I'll try to refactor my implementation based on this draft and I'll also try to address the other two comments (`parameterElements.push` and `Reference Object`).

char0n: Couple of more notes:

`(element.parent as MemberElement).value` - this is guaranteed by the refractor layer to always be `PathItemElement` so no asserts are needed here.

`return !test(resolvedPathTemplate, { strict: true });` - this is guaranteed to work because `openapi-path-templating` already supports that in latest version. 

image

If I'll think of something else I'll add comments here.

char0n: Related to `Reference Object` - if we found out that the linting is performed on non-dereferenced structure (which might be the case on 90%)  I can see following options:

### Option 1

Count the number of remaining template expressions and the the count of `Reference Objects`. If the count equals we just return true. It's hacky, but it's better than nothing. The corner case is that two distinct `Reference Objects` might point to the same `Parameter Object`. 

### Option 2

If there is a `Reference Object` within our collected parameters, we just don't do the validation at all and return `true`. 

---

For first iteration I would just do for **Option 2**.

kowalczyk-krzysztof:
@char0n I changed my implementation to be more semantic. I also modified the fixture to include a reference object and it seems that the linter function works on dereferenced specification so we don't need to handle that.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1011,35 @@ export const standardLinterfunctions: FunctionItem[] = [
...
-        return true;
+        const pathItemElement = (element.parent as MemberElement).value as ObjectElement;
+
+        if (pathItemElement.length === 0) {
+          return true;
+        }
+
+        const pathTemplateResolveParams: { [key: string]: string } = {};
+
+        pathItemElement.forEach((operation) => {
```
char0n, line 1022: `Path Item Object` can contain additional fixed fields like `servers`, `summary` or `description`. Which means the callback will receive `StringElement` (summary or description fields) or `ArrayElement` (servers field), and not just `OperationElement`. That's why I did following filtering in my "semantic code template":

```ts
        // we're asserting OperationElement polymorphically as this function has to support all OpenAPI versions
        const operationElements = Array.from(
          pathItemElement.filter((elm: Element) => elm.element === 'operation').map((elm) => elm),
        );
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1011,35 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          const parameters = (operation as OperationElement).get('parameters');
```
char0n, line 1023: This will throw error if `operation` is not `OperationElement` as described above.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1011,35 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+              if (isParameterElement(parameter)) {
+                const allowedLocation = ['path', 'query'];
+                if (allowedLocation.includes(toValue(parameter.in))) {
+                  pathTemplateResolveParams[toValue(parameter.name) as string] = 'placeholder';
+                }
+              }
+            });
+          }
+        });
+
```
char0n, line 1035: Code is not collecting `ParamterElements` for `PathItemElement` at all, as I've described in "semantic code template":

```ts
        // we're treating PathItemElement polymorphically as ObjectElement as this function has to support all OpenAPI versions
        const pathItemElement = (element.parent as MemberElement).value as ObjectElement;
        if (isArrayElement(pathItemElement.get('parameters'))) {
          parameterElements.push([...pathItemElement.get('parameters')]);
        }
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1011,35 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          return true;
+        }
+
+        const pathTemplateResolveParams: { [key: string]: string } = {};
+
+        pathItemElement.forEach((operation) => {
+          const parameters = (operation as OperationElement).get('parameters');
+          if (isArrayElement(parameters)) {
+            parameters.forEach((parameter) => {
+              if (isParameterElement(parameter)) {
```
char0n, line 1026: We cannot use semantic predicates here. This linting rule is not specific to any particular OpenAPI version, but rather is general to all OpenAPI versions (2.0, 3.0, 3.1). `isParameterElement` is specific to OpenAPI 3.0.x namespace and will not detect `ParamterElement` from OpenAPI 2.0.

We need to polymorphic check which will look like this:

```js
if (parameter.element === 'parameter') {
```

packages/apidom-ls/test/fixtures/validation/oas/path-template-all-defined.yaml
```diff
@@ -0,0 +1,70 @@
+openapi: 3.0.0
```
char0n, line 1: We also need test for OpenAPI 2.0 and OpenAPI 3.1.

packages/apidom-ls/test/fixtures/validation/oas/path-template-all-defined.yaml
```diff
@@ -0,0 +1,70 @@
...
+  parameters:
+    test_id:
+      name: test_id
+      in: path
+      required: true
+      schema:
+        type: string
+        format: uuid
+        title: Test Id
+paths:
```
char0n, line 15: The fixture is missing cases for `PathItem.parameters` - when parameters are defined only in `PathItem.paramters` or both in `PathItem.parameters` and `Operation.parameters`.

char0n (changes requested):
We're getting there! Please have a look at my CR comments. I'll duble-check the `Reference Object` handling when we have current CR comments handled.

kowalczyk-krzysztof:
Thanks for the comments. I misunderstood the structure of `Path Item Object` and made some wrong assumptions 🤦🏻 But now I see what I've missed.
I'll also add fixtures and tests for OAS2.0 and OAS3.1 and modify the OAS3.0 one to take care of the missing cases.

> We cannot use semantic predicates here.
Yeah I see, Intellisense was showing me a lot of different import options and I got confused and I choose one at random, I wasn't aware that those predicate functions differ. I also didn't know about the polymorphic check option.

kowalczyk-krzysztof:
I modified my implementation according to above comments and I added fixtures for different OAS versions.
I'm not sure how to get rid of those `@ts-ignore` though since I can't really make assertions as there are no "generic" (as in, valid for any OAS version) types.

char0n (commented):
Some final nitpicks and possible enhancements.

char0n:
> I'm not sure how to get rid of those @ts-ignore though since I can't really make assertions as there are no "generic" (as in, valid for any OAS version) types.

We can use lowest common denominator which is `ObjectElement`. It should resolve issues with @ts-ignore

kowalczyk-krzysztof:
Pushed the suggested changes. Thanks for the `get` suggestion, I completely forgot it exists. However, this suggestion:
```typescript
parameterElements.forEach((parameter: ObjectElement) => {
  if (allowedLocation.includes(toValue(parameter.get('in')))) {
    pathTemplateResolveParams[toValue(parameter.get('name')) as string] = 'placeholder';
  }
});
``` 
doesn't work unfortunately, I still need to cast `parameter` (can't assign a type here) and changing the line below to be `ObjectElement[]`
```typescript
const parameterElements: Element[] = [];
```
requires type casting in other places so I think this:
```typescript
 parameterElements.forEach((parameter) => {
          if (allowedLocation.includes(toValue((parameter as ObjectElement).get('in')))) {
            pathTemplateResolveParams[toValue((parameter as ObjectElement).get('name'))] =
              'placeholder';
          }
        });
```
is the cleanest looking solution.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1010,55 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+
+        if (pathItemElement.length === 0) {
+          return true;
+        }
+
+        const parameterElements: Element[] = [];
+        const isParameterElement = (el: Element): boolean => el.element === 'parameter';
+
+        pathItemElement.forEach((el) => {
+          if (el.element === 'parameters') {
```
char0n, line 1023: Just found this **bug**. This will never be true. You can try creating a failing test for this. Basically there is no ApiDOM element with `parameters` type. `PathItem.parameters` is an `ArrayElement` with `path-item-paramters` meta class. What you're interested here is asserting on `PathItemParametersElement` NCE (Non Concrete Element type).

NCEs:
 - OpenAPI 2.0 - https://github.com/swagger-api/apidom/blob/main/packages/apidom-ns-openapi-2/src/elements/nces/PathItemParameters.ts
 - OpenAPI 3.x.y - https://github.com/swagger-api/apidom/blob/main/packages/apidom-ns-openapi-3-0/src/elements/nces/PathItemParameters.ts

```suggestion
          if (includesClasses(['path-item-parameters'], el)) {
```

---

`includesClasses` is export from `@swagger-api/apidom-core` package.

char0n: But at this point IMHO it make more sense to pull the first `if` statement outside of the loop and just process it manually as I did in the "semantic code template". You will avoid asserting for meta class on every `Path Item` fixed field and converting meta classes to POJOS (done by `includesClasses`).

kowalczyk-krzysztof: Right. I initially did it that way but then I put the check inside loop as I thought it looked cleaner and I wasn't aware of the implications of it. 
Changed.

kowalczyk-krzysztof:
> All right, we're looking good! Nice job!
> 
> **WARNING**: please don't merge, I'll need to double check the `Reference Object` handling and verify that we're working on dereferenced structure. After that I'll do the "sqash and merge".
> 
> Thanks for your patience ;]

Thanks for all the comments and suggestions 😄 
I learned a lot about ApiDOM in this PR and hopefully my future PRs will require less help.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1009,9 +1010,67 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        const resolveResult = !test(resolvedPathTemplate, { strict: true });
+
+        if (!resolveResult && oneOfParametersIsReferenceObject) {
+          return true;
+        }
+        return resolveResult;
       }
-      return true;
+
+      return false;
```
char0n, line 1073: This must return true always as it signals skipping the rule.

```suggestion
      return true;
```

kowalczyk-krzysztof: Oops 😅 Fixed.

Issue #3549, opened 2023-12-18
JSON Schema: create linting rule detecting unknown keyword


char0n:
The rule should create `WARNING` diagnostic severity and should fail on every JSON Schema that is of `object type`, contains non JSON Schema keywords and doesn't contain any JSON Schema keywords.

PR #3554, opened 2023-12-18
feat(apidom-ls): create schema rule for missing core keywords

frantuma:
Refs #3549

Also fixes rules for JSON Schema type checks considering also boolean schemas


packages/apidom-ls/src/config/asyncapi/target-specs.ts
```diff
@@ -0,0 +1,18 @@
...
+export const AsyncAPI2 = [
+  ...AsyncAPI200,
+  ...AsyncAPI210,
+  ...AsyncAPI220,
+  ...AsyncAPI230,
+  ...AsyncAPI240,
+  ...AsyncAPI250,
+  ...AsyncAPI260,
+];
+export const AsyncAPI = [...AsyncAPI2];
```
char0n, line 18: In light of https://github.com/swagger-api/apidom/commit/1e15838314e7b2eea7ffa4d8259e25f41b9a2d64 can we remove this?

The reasoning is that we'd rather add target spec explicitly to what applies, then automatically activating existing rules for example for AsyncAPI 3.0 (when we're adding support for it). This total group of all AsyncAPI spec targets has no practical usage and causes more issues than it solves.

packages/apidom-ls/src/config/common/schema/lint/additional-items--type.ts
```diff
@@ -9,7 +9,7 @@ const additionalItemsTypeLint: LinterMeta = {
   message: 'additionalItems must be a schema',
   severity: DiagnosticSeverity.Error,
   linterFunction: 'apilintElementOrClass',
-  linterParams: ['schema'],
+  linterParams: ['schema', 'boolean'],
```
char0n, line 12: We cannot really do that. OpenAPI 2.0/3.0.x use JSON Schema Draft 4/5 which doesn't support `Boolean JSON Schema`.  Boolean JSON Schemas are support since JSON Schema Draft 6 and applies to: AsyncAPI 2.x and OpenAPI 3.1.0 only.

We would have to fork this rule and use `targetSpec` to apply it to appropriate specification and their versions.

> [!IMPORTANT]  
> This applies to all modified rules in `/common/schema/`

packages/apidom-ls/src/config/common/schema/lint/missing-core-fields-oas-3-0-asyncapi.ts
```diff
@@ -0,0 +1,91 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
```
char0n, line 1: Just to make naming consistent with the import symbol:

`missing-core-fields-oas-3-0-asyncapi.ts` -> `missing-core-fields-openapi-3-0-asyncapi-2.ts`

packages/apidom-ls/src/config/common/schema/lint/missing-core-fields-oas-3-0-asyncapi.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+import { OpenAPI3 } from '../../../openapi/target-specs';
+import { AsyncAPI } from '../../../asyncapi/target-specs';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const missingCoreFieldsOpenAPI3_0AndAsyncAPILint: LinterMeta = {
```
char0n, line 9: IMHO this needs to be forked to multipe rules targeting different specs.

OpenAPI 2.0/3.0.x uses JSON Schema Draft 4/5 with following keywords: https://github.com/swagger-api/apidom/blob/main/packages/apidom-ns-json-schema-draft-4/src/elements/JSONSchema.ts#L35

AsyncAPI 2.x uses JSON Schema Draft 7 which:
Removes keywords: id
Add keywords: $id, contains, propertyNames, examples, $comment, if, then, else, contentEncoding, contentMediaType, media, writeOnly

OpenAPI 3.1 uses JSON Schema Draft 2020-12 with following keywords: https://github.com/swagger-api/apidom/blob/5b504e523c532a31aaf1b133a14ff3b76c126435/packages/apidom-ns-openapi-3-1/src/elements/Schema.ts

char0n (changes requested):
Looks good! We need to sort some cross specs details before we go for a merge.

frantuma:
replaced by https://github.com/swagger-api/apidom/pull/3640

Issue #3104, opened 2023-08-31
OpenAPI 2.0 - produce linting/docs/completion rules in apidom-ls


char0n:
The goal of this issue is to produce linting/docs/completion rules.

**TODO**:

- [x] Documentation rules
- [x] Completion rules
- [x] Linting rules

char0n:
Remaining rules have been ticketed into individual GitHub issues.

PR #3561, opened 2023-12-19
feat(ls): add rules for OpenAPI 2.0 Operation Object

char0n:
Refs #3104


packages/apidom-ls/src/config/openapi/operation/lint/schemes--type.ts
```diff
@@ -0,0 +1,20 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+import { OpenAPI2 } from '../../target-specs';
+
+const schemesTypeLint: LinterMeta = {
+  code: ApilintCodes.OPENAPI2_OPERATION_FIELD_SCHEMES_TYPE,
+  source: 'apilint',
+  message: "'consumes' must be an array of strings",
```
char0n, line 10:
```suggestion
  message: "'schemes' must be an array of strings",
```

Issue #3565, opened 2023-12-20
OpenAPI: Swagger 2.0 basePath linting rule fails when value contains 's'


CalemRoelofsSB:
### Q&A (please complete the following information)
 - OS: MacOS Ventura 13.6.3
 - Browser: Firefox 
 - Version: 121.0
 - Method of installation: found while using https://editor-next.swagger.io/
 - ApiDOM version: 0.85.0
 - Swagger/OpenAPI version: Swagger 2.0

### Describe the bug you're encountering
The Swagger 2.0 linting rule for validating that the `basePath` is correctly formatted (https://github.com/swagger-api/apidom/edit/main/packages/apidom-ls/src/config/openapi/swagger/lint/base-path--pattern.ts) fails erroneously whenever there is a `basePath` with the character `s` in it.

### To reproduce...

1. Load up any Swagger 2.0 definition in https://editor-next.swagger.io/
2. Add `basePath: /s123` or `basePath: /123s` or `basePath: /hello/worlds` or any string starting with `/` and containing the character `s`

### Expected behavior
No validation error is displayed

### Screenshots
![image](https://github.com/swagger-api/apidom/assets/96050034/7d7d15c5-6266-42d9-a8cb-94b2ca9f35bf)
![image](https://github.com/swagger-api/apidom/assets/96050034/07ed5342-96b0-4b69-8a55-5604575a7f06)
![image](https://github.com/swagger-api/apidom/assets/96050034/359b9714-b00b-46b5-a1d9-15a15dc229cd)

### Additional context or thoughts
This appears to be caused by the regex used for this rule found [here](https://github.com/swagger-api/apidom/edit/main/packages/apidom-ls/src/config/openapi/swagger/lint/base-path--pattern.ts)

char0n:
Addressed in https://github.com/swagger-api/apidom/pull/3566

PR #3566, opened 2023-12-20
fix(ls): Fix regex in Swagger 2.0 basePath linting rule

CalemRoelofsSB:
Refs #3565 

This PR fixes the regex in the `OPENAPI2_SWAGGER_FIELD_BASE_PATH_PATTERN` linting rule so that it matches on whitespace characters instead of the letter 's'

### DIFF
--- a/packages/apidom-ls/src/config/openapi/swagger/lint/base-path--pattern.ts
+++ b/packages/apidom-ls/src/config/openapi/swagger/lint/base-path--pattern.ts
@@ -10,7 +10,7 @@ const basePathPatternLint: LinterMeta = {
   message: '"basePath" value MUST be a relative URI Referencing starting with a leading slash (/).',
   severity: DiagnosticSeverity.Error,
   linterFunction: 'apilintValueRegex',
-  linterParams: ['^/(?:[^/s][^s]*)?$'],
+  linterParams: ['^/(?:[^/\s][^\s]*)?$'],
   target: 'basePath',
   marker: 'value',
   data: {},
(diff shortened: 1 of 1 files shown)


CalemRoelofsSB:
It's interesting that prettier sees these as useless escapes https://github.com/swagger-api/apidom/actions/runs/7274733595/job/19821257422?pr=3566

packages/apidom-ls/src/config/openapi/swagger/lint/base-path--pattern.ts
```diff
@@ -10,7 +10,7 @@ const basePathPatternLint: LinterMeta = {
   message: '"basePath" value MUST be a relative URI Referencing starting with a leading slash (/).',
   severity: DiagnosticSeverity.Error,
   linterFunction: 'apilintValueRegex',
-  linterParams: ['^/(?:[^/s][^s]*)?$'],
+  linterParams: ['^/(?:[^/\s][^\s]*)?$'],
```
char0n, line 13:
```suggestion
  linterParams: ['^/(?:[^/\\s][^\\s]*)?$'],
```

char0n (changes requested):
Hi @CalemRoelofsSB, suggested the change that actually works.

By doing single `\s` we're actually putting `s` character there again. We need to do double slashing to actually create a `\s` string: `\\s`.

CalemRoelofsSB:
@char0n ah perfect, sorry about that! I've added it in now.

char0n (approved):
Nice catch! Thank you!

char0n:
@CalemRoelofsSB plese next time squash + merge your PR or let me do it. This PR created 3 commit messages in main branch and broke the linear history.

Issue #3546, opened 2023-12-18
OpenAPI: linting rule for checking if parameter is defined within the path template


char0n:
Originally defined as **Rule 2** in https://github.com/swagger-api/apidom/issues/3517

We want a linting rule that validates that user did not forget to define template expressions for all the appropriate Parameter Objects.

This linting rule is going to be attached to `Parameter Object` and not to `path-template` meta class so that the `Parameter Objects` gets highlighted in red if linting rule fails.

This is going to be a bit tricky as `Parameter Objects` first needs to establish if it has been defined for `Path Item Object` or `Operation Object` that is attached to `Path Item Object` via `http-verb`. 

Next it needs to check if it `location` is within `path` or `query`.

Next is needs to check if the `path template` attached to it's `Path Item Object` contains the template expression matching `Parameter.name`.

char0n:
All right, so our first order of business if to determine if the `Parameter Object` we're looking at is somehow related to `Path Item Object` that is attached inside the `Paths Object`. Let's focus on that first, and after we have this established we'll focus on looking at `Operation Object` parameters.

### Path Item Object parameters

We're specifically talking about `PathItem.parameters`. 

#### 1. Determining Path Item relationship

We first need to determine if `Parameter Object` is part of `PathItem.parameters`. That is achieved by looking at `Parameter Object` parent and checking if the parent is `ArrayElement` with meta class of `path-item-parameters`. 

Predicate will look like this:

```js
const isInPathItemElement = isArrayElement(paramerElement.parent) && includesClasses(['path-item-parameters'], parameterElement.parent)
const pathItemElement = parameterElement.parent.parent.parent
```

> [!TIP]
> We can avoid step **1.** and all the following steps by immediately looking into `Parameter.in` values and checking if it's `path`. 

#### 2. Determining if Path Item is part of path templating

To determine if the found `Path Item Object` (the one we found in previous step) is part of path templating, we have to look at it's meta field called `path`. If `Path Item Object` has this meta field and the meta field is of `StringElement` type, that we've determined that `Parameter Object` is part of path templating.

Predicate will look like this:

```js
const isPathItemPartOfPathTemplating = isStringElement(pathItemElement.meta.get('path');
const pathTemplate = toValue(pathItemElement.meta.get('path'));
```

#### 3. Determining if Parameter Object is matched with path templating

[...]

PR #3571, opened 2023-12-22
feat(apidom-ls): lint parameter defined in path template

kowalczyk-krzysztof:
Refs #3546


char0n:
All right, so our first order of business if to determine if the `Parameter Object` we're looking at is somehow related to `Path Item Object` that is attached inside the `Paths Object`. Let's focus on that first, and after we have this established we'll focus on looking at `Operation Object` parameters.

### Path Item Object parameters

We're specifically talking about `PathItem.parameters`. 

#### 1. Determining Path Item relationship

We first need to determine if `Parameter Object` is part of `PathItem.parameters`. That is achieved by looking at `Parameter Object` parent and checking if the parent is `ArrayElement` with meta class of `path-item-parameters`. 

Predicate will look like this:

```js
const isInPathItemElement = isArrayElement(paramerElement.parent) && includesClasses(['path-item-parameters'], parameterElement.parent)
const pathItemElement = parameterElement.parent.parent.parent
```

> [!TIP]
> We can avoid step **1.** and all the following steps by immediately looking into `Parameter.in` values and checking if it's one of `path` or `query`. 

#### 2. Determining if Path Item is part of path templating

To determine if the found `Path Item Object` (the one we found in previous step) is part of path templating, we have to look at it's meta field called `path`. If `Path Item Object` has this meta field and the meta field is of `StringElement` type, that we've determined that `Parameter Object` is part of path templating.

Predicate will look like this:

```js
const isPathItemPartOfPathTemplating = isStringElement(pathItemElement.meta.get('path');
const pathTemplate = toValue(pathItemElement.meta.get('path'));
```

#### 3. Determining if Parameter Object is matched with path templating

This is a bit tricky. We'll have to use [openapi-path-templating](https://github.com/char0n/openapi-path-templating) to parse the path template into the AST:

This path template `/pets?offset{offset}limit={limit}` will parse into following AST:

```
          ['path-template', '/pets?offset{offset}limit={limit}'],
          ['path', '/pets'],
          ['slash', '/'],
          ['path-literal', 'pets'],
          ['query-marker', '?'],
          ['query', 'offset{offset}limit={limit}'],
          ['query-literal', 'offset'],
          ['template-expression', '{offset}'],
          ['template-expression-param-name', 'offset'],
          ['query-literal', 'limit='],
          ['template-expression', '{limit}'],
          ['template-expression-param-name', 'limit'],
```

If the `Parameter.in` === `path` - we'll have to look for `template-expression-param-name` with value of `Parameter.name` inside the AST but match must come **before** the `query-marker` AST node type.

If the `Parameter.in` === `query` - we'll have to look for `template-expression-param-name` with value of `Parameter.name` inside the AST but match must come **after** the `query-marker` AST node type.


--- 

Only after we have the above implementation, it makes sense to focus on `Operation.parameters`...

kowalczyk-krzysztof:
Thanks, for the comment. I made some changes and it seems to be working fine for `Operation.parameters` but I'm still stuck on when the parameter is defined like this:
```yaml
openapi: 3.0.0
info:
  title: Foo
  version: 0.1.0
components:
  parameters:
    test_id:
      name: test_id
      in: path
      required: true
      schema:
        type: string
        format: uuid
        title: Test Id
```
What do I do in this case?

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,52 @@ export const standardLinterfunctions: FunctionItem[] = [
       return true;
     },
   },
+  {
+    functionName: 'apilintOpenAPIParameterFieldIsDefinedWithinPathTemplate',
+    function: (element: Element) => {
+      if (element.element === 'parameter') {
+        const parameterName = toValue((element as ObjectElement).get('name'));
+        const parameterLocation = toValue((element as ObjectElement).get('in'));
+        const isInPathItemElement =
```
char0n, line 1078: We can bail out right here, if it's not a path parameter:

```js
        if (parameterLocation !== 'path') {
          return true;
        }
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,52 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          isArrayElement(element.parent) &&
+          includesClasses(['path-item-parameters'], element.parent);
+        const isChildOfOperationElement = element.parent.parent.parent.element === 'operation';
```
char0n, line 1081: Let's not deal with `Operation.parameters` at this point of time. It will just unnecessary complicate the implementation.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,52 @@ export const standardLinterfunctions: FunctionItem[] = [
       return true;
     },
   },
+  {
+    functionName: 'apilintOpenAPIParameterFieldIsDefinedWithinPathTemplate',
+    function: (element: Element) => {
+      if (element.element === 'parameter') {
+        const parameterName = toValue((element as ObjectElement).get('name'));
+        const parameterLocation = toValue((element as ObjectElement).get('in'));
```
char0n, line 1077: We can bail out right here, if it's not a path parameter:

```js
        if (parameterLocation !== 'path') {
          return true;
        }
```

packages/apidom-ls/src/services/validation/linter-functions.ts
char0n, line 1081: We can bail out right here, if the parameter is not part of `PathItem.parameters`:

```js
if (!isInPathItemElement) {
  return true;
}
```

char0n: Next we need to estabilish the `pathItemElement`:

```js
const pathItemElement = parameterElement.parent.parent.parent
```

char0n: Next we need to estabilish if `Path Item Object` is part of path templating (inside Paths Object) and not part of Webhook, Callback or Components objects:

```js
const isPathItemPartOfPathTemplating = isStringElement(pathItemElement.meta.get('path');
```

char0n: We can bail out right here, if `Path Item Object` is **not** part of path templating:

```js
if (!isPathItemPartOfPathTemplating) {
  return true
}
```

char0n: After the above check, we can actually do the lookup of parameter name inside the path template.

kowalczyk-krzysztof: Ok. I'll remove the code for that.

packages/apidom-ls/src/services/validation/linter-functions.ts
char0n: > Just the path check isn't enough as Parameter objects defined in components can also have in with value path and it passes this check.

We're not interested in those Parameter Objects as they are not used in the process of path templating. And yes it should pass the check (return true).

>  I'm not sure it's valid OpenAPI definition, but if it is, it has to be accounted for.

It's valid OpenAPI definition, and it's accounted for with checks mentioned in previous comment.

kowalczyk-krzysztof: > It's valid OpenAPI definition, and it's accounted for with checks mentioned in https://github.com/swagger-api/apidom/pull/3571#discussion_r1440635491.

Yeah I was just trying to say that the checko below:
```javascript
 if (parameterLocation !== 'path') {
     return true;
  }
``` 
Isn't enough on it's own. 

> We can avoid step 1. and all the following steps by immediately looking into Parameter.in values and checking if it's path.

The above tip from the initial comment confused me a bit and I thought for a while that just this check would be enough, but as discussed, it isn't.

char0n: No worries, this is one of the most complex linting function. We have to be synced with how we want to approach the problem and why.

If something is ambiguous in https://github.com/swagger-api/apidom/pull/3571#issuecomment-1873771768, we can have a zoom call and sync.

char0n:
All right, so as discussed we'll handle the `Operation.parameters` in follow-up PR.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,41 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+  {
+    functionName: 'apilintOpenAPIParameterFieldIsDefinedWithinPathTemplate',
+    function: (element: Element) => {
+      if (element.element === 'parameter') {
+        const parameterLocation = toValue((element as ObjectElement).get('in'));
+        const isInPathItemElement =
+          isArrayElement(element.parent) &&
+          includesClasses(['path-item-parameters'], element.parent);
+
+        if (!isInPathItemElement || parameterLocation !== 'path') {
```
char0n, line 1081: Just an optimization tip (which is up to you to utilize):

if you bail out directly after:

```js
const parameterLocation = toValue((element as ObjectElement).get('in'));

if (parameterLocation !== 'path') {
  return true;
}
```

You'll avoid `isInPathItemElement` assertions which does perform some ApiDOM conversions under the hood which might slightly hit performance.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,41 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        }
+
+        const pathTemplate = toValue(pathItemElement.meta.get('path'));
+        const parameterName = toValue((element as ObjectElement).get('name'));
+
+        const parseResult = parse(pathTemplate);
+        const parts: [string, string][] = [];
+        parseResult.ast.translate(parts);
+
+        const pathTemplateASTIncludesParameter = (ast: [string, string][]) =>
```
char0n, line 1099: Do we need to define an ad-hoc function here? It looks like a lot of code that can be simplified a bit. Here is a suggestion:

```js
const parseResult = parse(pathTemplate);
const parts: [string, string][] = [];

parseResult.ast.translate(parts);
return parts.some(([name, value]) => name === 'template-expression-param-name' && value === parameterName);
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,41 @@ export const standardLinterfunctions: FunctionItem[] = [
       return true;
     },
   },
+  {
+    functionName: 'apilintOpenAPIParameterFieldIsDefinedWithinPathTemplate',
```
char0n, line 1073: What would you say for the following simplified name?

```suggestion
    functionName: 'apilintOpenAPIParameterInPathTemplate',
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1068,4 +1069,41 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        const isPathItemPartOfPathTemplating = isStringElement(pathItemElement.meta.get('path'));
+
+        if (!isPathItemPartOfPathTemplating) {
+          return true;
+        }
+
+        const pathTemplate = toValue(pathItemElement.meta.get('path'));
+        const parameterName = toValue((element as ObjectElement).get('name'));
+
+        const parseResult = parse(pathTemplate);
```
char0n, line 1095: We're missing check if the `path template` is valid:

```suggestion
        const parseResult = parse(pathTemplate);
        if (!parseResult.result.success) return true;
```

packages/apidom-ls/src/config/openapi/parameter/lint/is-defined-within-path-template.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const isDefinedWithinPathTemplate: LinterMeta = {
+  code: ApilintCodes.OPENAPI2_PARAMETER_FIELD_IS_DEFINED_WITHIN_PATH_TEMPLATE,
+  source: 'apilint',
+  message: 'parameter is not defined within path template',
+  severity: DiagnosticSeverity.Error,
+  linterFunction: 'apilintOpenAPIParameterFieldIsDefinedWithinPathTemplate',
+  marker: 'value',
+  targetSpecs: [...OpenAPI2, ...OpenAPI3],
+  conditions: [
```
char0n, line 15: No need for this condition; This condition checks for valid path template, but Parameter Object has no direct connection to the path templating. (connection is via Path Item Object).


It is handled directly in the linting function: https://github.com/swagger-api/apidom/pull/3571/files#r1441457612

packages/apidom-ls/src/config/openapi/parameter/lint/is-defined-within-path-template.ts
```diff
@@ -0,0 +1,23 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes';
+import { LinterMeta } from '../../../../apidom-language-types';
+import { OpenAPI2, OpenAPI3 } from '../../target-specs';
+
+const isDefinedWithinPathTemplate: LinterMeta = {
```
char0n, line 7:
```suggestion
const definedWithinPathTemplate: LinterMeta = {
```

char0n: We're not using `is/are/has` prefixes with the predicate naming by convention. Example:

we use `required-fields` instead of `are-required-fields`.

packages/apidom-ls/src/config/codes.ts
```diff
@@ -734,6 +734,7 @@ enum ApilintCodes {
   OPENAPI2_PARAMETER_FIELD_UNIQUE_ITEMS_TYPE = 3101700,
   OPENAPI2_PARAMETER_FIELD_ENUM_TYPE = 3101800,
   OPENAPI2_PARAMETER_FIELD_MULTIPLE_OF_TYPE = 3101900,
+  OPENAPI2_PARAMETER_FIELD_IS_DEFINED_WITHIN_PATH_TEMPLATE = 3102000,
```
char0n, line 737:
```suggestion
  OPENAPI2_PARAMETER_FIELD_DEFINED_IN_PATH_TEMPLATE = 3102000,
```

packages/apidom-ls/src/config/openapi/parameter/lint/is-defined-within-path-template.ts
char0n: Name of the file `defined-in-path-template.ts`

kowalczyk-krzysztof:
@char0n Thanks for the suggestions. I implemented all of them and I also updated fixtures to include all the possible places where `Parameters` object can be defined.

char0n (approved):
Looking good here. Nice job!

PR #3629, opened 2024-01-04
feat(apidom-ls): add operations to parameter lint

kowalczyk-krzysztof:
refs #3546 Operation Object parameters

### DIFF
--- a/packages/apidom-ls/src/services/validation/linter-functions.ts
+++ b/packages/apidom-ls/src/services/validation/linter-functions.ts
@@ -1087,15 +1087,11 @@ export const standardLinterfunctions: FunctionItem[] = [
 
         if (!isInPathItemElement && !isInOperationElement) return true;
 
-        let pathItemElement = {} as Element;
+        const pathItemElement: Element | undefined = isInOperationElement
+          ? element.parent?.parent?.parent?.parent?.parent
+          : element.parent?.parent?.parent;
 
-        if (isInOperationElement) {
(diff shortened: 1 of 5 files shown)


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1081,9 +1081,22 @@ export const standardLinterfunctions: FunctionItem[] = [
...
           includesClasses(['path-item-parameters'], element.parent);
 
-        if (!isInPathItemElement) return true;
+        const isInOperationElement =
+          isArrayElement(element.parent) &&
+          includesClasses(['operation-parameters'], element.parent);
+
+        if (!isInPathItemElement && !isInOperationElement) return true;
+
+        let pathItemElement = {} as Element;
```
char0n, line 1090: Would it be possible to simplify into following form?

```js
const pathItemElement: Element | undefined = isInOperationElement
  ? element.parent?.parent?.parent?.parent?.parent
  : element.parent?.parent?.parent;

if (pathItemElement?.element !== 'pathItem') return true;
```

NOTE that I've also made traversing the parents safe by optional chaining, which should eliminate future corner cases.

kowalczyk-krzysztof: Yep it works fine 👍🏻 I changed the implementation.

char0n (approved):
LGTM!

Issue #3481, opened 2023-11-30
Replace stampit with TypeScript mixins


char0n:
The goal of this issue is to determine if it make sense to switch from stamps to mixins.  If it is determined that we're going to do that, another issue should be created to track the implementation progress there.

Pragmatically it just make sense to be able to use advanced TypeScript 5 features like decorators. Let's push through with this on OpenAPI 2.0 namespace, and see where we'll get.

---

**TODO**:

- [x] Documentation changes

---

**Resources:**
- https://medium.com/@koresar/fun-with-stamps-episode-16-typescript-mix-in-classes-vs-stamps-5e3c548753ea
- https://www.typescriptlang.org/docs/handbook/mixins.html

char0n:
Additional notes:

After further research I've come into the conclusion (**subjective one**), that It's currently really not worth the effort. If we would like to switch to TypeScript mixins we would have to rewrite significant amount of code and we'd still would have to use some kind of abstraction like [ts-mixer](https://www.npmjs.com/package/ts-mixer) to provide convenient API.

We would replace one way of doing things (IMHO more flexible with stamps) with another one, which is less flexible.

char0n:
Another package that has been identified is https://github.com/swagger-api/apidom/tree/main/packages/apidom-ast

All the AST node types can be represented as TypeScript classes as there isn't multiple inheritance involved.

glowcloud:
The packages that I've been working on:
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser-adapter-json
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser-adapter-yaml-1-2
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-ast

I'll make the suggested changes from https://github.com/swagger-api/apidom/pull/3646 to Parser and to the other packages, as some cases will apply there as well. The Adapter packages should be ready for a PR then. After that, I'll continue the work on the AST package.

I believe that these packages can be changed as well:
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-core
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-reference
- https://github.com/swagger-api/apidom/tree/main/packages/apidom-ls - only one case here

glowcloud:
Status update for each of the previously mentioned packages:

Changes merged into main:
[...]

PR #3646, opened 2024-01-09
refactor(parser): change from stamps to classes

glowcloud:
Refs #3481


packages/apidom-parser/src/types.d.ts
```diff
@@ -19,7 +19,7 @@ export interface ApiDOMParserAdapter {
 
 export interface ApiDOMParser {
```
char0n, line 20: No need for this interface anymore.

When you define
```ts
class ApiDOMParser {
...
}
```
then this code automatically generates instance type and class type in `type scope`.

**Proof**

```ts
type InstanceType = ApiDOMParser;
type ClassType = typeof ApiDOMParser;
```

packages/apidom-parser/src/parser.ts
```diff
@@ -18,119 +17,124 @@ export type {
 
 export { ParserError };
 
-const ApiDOMParser: stampit.Stamp<ApiDOMParserType> = stampit().init(
-  function ApiDOMParserConstructor() {
-    const adapters: ApiDOMParserAdapter[] = [];
+class ApiDOMParserClass implements ApiDOMParserType {
```
char0n, line 20:
```suggestion
class ApiDOMParser {
```
Explanation in https://github.com/swagger-api/apidom/pull/3646/files#r1445972385

packages/apidom-parser/src/parser.ts
```diff
@@ -18,119 +17,124 @@ export type {
...
-      return candidates;
-    };
+    return candidates;
+  }
 
-    const findAdapter = async (source: string, mediaType: string | undefined) => {
-      if (isString(mediaType)) {
-        return adapters.find((adapter) => {
-          if (!isArray(adapter.mediaTypes)) return false;
+  private async findAdapter(
```
char0n, line 36: Let's allow users to make subclasses and reuse the logic.

```suggestion
  protected async findAdapter(
```

packages/apidom-parser/src/parser.ts
```diff
@@ -18,119 +17,124 @@ export type {
...
-        throw new ParserError(
-          'Encountered an unexpected error while matching parser adapters against the source.',
-          {
-            source,
-            parserOptions: options,
-            cause: error,
-          },
-        );
-      }
+    // @ts-ignore
```
char0n, line 97: IMHO we will be able to remove this already.
```suggestion
    // @ts-ignore
```

packages/apidom-parser/package.json
```diff
@@ -42,8 +42,7 @@
     "@swagger-api/apidom-error": "^0.90.0",
     "@types/ramda": "~0.29.6",
     "ramda": "~0.29.1",
-    "ramda-adjunct": "^4.1.1",
-    "stampit": "^4.3.2"
+    "ramda-adjunct": "^4.1.1"
```
char0n, line 45: When you remove or add dependency from particular package, you need to run `npm i` in top level monorepo. This will regenerate package-lock.json.

packages/apidom-parser/src/parser.ts
```diff
@@ -18,119 +17,124 @@ export type {
...
+    const adapter = await this.findAdapter(source, options.mediaType);
 
-      return adapter?.namespace;
-    };
+    return adapter?.namespace;
+  }
 
-    this.findMediaType = async function findMediaType(source: string) {
-      const adapter = await findAdapter(source, undefined);
+  async findMediaType(source: string): Promise<string | void> {
```
char0n, line 67: Can you maybe explain reason why you used `Promise<string | void>` instead of `Promise<string | undefined>`? 

Here is my understanding:

> Promise<string | undefined> is more explicit in indicating that the resolved value of the Promise can be either a string or undefined. On the other hand, Promise<string | void> suggests that the focus is not on the resolved value, which could be a string or nothing significant.

packages/apidom-parser/src/parser.ts
```diff
@@ -18,119 +17,124 @@ export type {
...
+      throw new ParserError('Parsing encountered an unexpected error.', {
+        source,
+        parserOptions: options,
+        cause: error,
+      });
+    }
+  }
+}
+
+function ApiDOMParser(): ApiDOMParserClass {
```
char0n, line 136: We can directly define a constructor within the `ApiDOMParser` class and use [target.new](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target) meta-property to check for presence of `new` keyword.

char0n (changes requested):
Looks good! Please have a look at my CRCs (code review comments).

char0n (approved):
LGTM! Nice job!

PR #3681, opened 2024-01-16
refactor(ast): change from stamps to TypeScript classes

glowcloud:
Refs #3481


packages/apidom-ast/src/Node.ts
```diff
@@ -1,45 +1,42 @@
-import stampit from 'stampit';
-
 import Position from './Position';
 
-interface Node {
-  type: string | null;
-  isMissing: boolean;
-  children: unknown[];
-  position: Position | null;
+export interface NodeConstructor {
```
char0n, line 3: You don't need to do this anymore.

When you define Node class

**node.ts**
```ts
class Node {
}
export default Node;
```
It defines two interfaces for you under the hood

1. instance interface
2. constructor interface

**Accessing instance interface**
```ts
import Node from './node';

interface Test extends Node {
}
```

**Accessing constructor interface**
```js
import Node from './node';

type NodeConstructor = typeof Node;
```

char0n:
Take a look at following proposal of how the classes might look like. What do we exactly see here?

1. Constructor options are extracted into separate interface which can be extended 
2. `type` is defined both as public instance property and static class property
3. If any class extends `Node` and redefine it's static `type` property, the super() constructor of Node class will take care of assigning to instance
4. We can use `...rest` destructuring pattern in type safe way to avoid mentioning constructor options explicitly now that we use explicit interface for it

**Node.ts**
```ts
import Position from './Position';

export interface NodeOptions {
  children?: unknown[];
  position?: Position | null;
  isMissing?: boolean;
}

class Node {
  public static readonly type: string = 'node';

  public readonly type: string = 'node';

  public isMissing: boolean;

  public children: unknown[];

  public position: Position | null;

  constructor({ children = [], position = null, isMissing = false }: NodeOptions = {}) {
    this.type = (this.constructor as unknown as Node).type;
    this.isMissing = isMissing;
    this.children = children;
    this.position = position;
  }
}

export default Node;
```

**Error.ts**
```ts
import Node from './Node';
import type { NodeOptions } from './Node';

export interface ErrorOptions extends NodeOptions {
  value?: unknown;
  isUnexpected?: boolean;
}

class Error extends Node {
  public static readonly type = 'error';

  public value: unknown;

  public isUnexpected: boolean;

  constructor({ value = null, isUnexpected = false, ...rest }: ErrorOptions = {}) {
    super({ ...rest });
    this.value = value;
    this.isUnexpected = isUnexpected;
  }
}

export default Error;
```

packages/apidom-ast/src/json/nodes/JsonDocument.ts
```diff
@@ -1,23 +1,13 @@
...
-  },
-  methods: {
-    // @ts-ignore
-    get child(): unknown {
-      // @ts-ignore
-      return head(this.children);
-    },
-  },
-});
+  public get child(): unknown | null {
```
char0n, line 8: `unknown` should be just fine here as it already contains `null` and `undefined` in it's definition.

```suggestion
  public get child(): unknown | null {
```

packages/apidom-ast/src/yaml/nodes/YamlSequence.ts
```diff
@@ -1,35 +1,24 @@
...
-const YamlSequence: stampit.Stamp<YamlSequence> = stampit(YamlCollection, {
-  statics: {
-    type: 'sequence',
-  },
-  propertyDescriptors: {
-    content: {
-      get(): Array<YamlSequence | YamlMapping | YamlScalar | YamlAlias> {
-        // @ts-ignore
-        const { children } = this;
+Object.defineProperty(YamlSequence, 'content', {
```
char0n, line 11:
```suggestion
Object.defineProperty(YamlSequence.prototype, 'content', {
```

packages/apidom-ast/src/yaml/nodes/YamlStream.ts
```diff
@@ -1,32 +1,19 @@
...
-      get(): Array<YamlDocument | YamlComment> {
-        // @ts-ignore
-        return Array.isArray(this.children)
-          ? // @ts-ignore
-            this.children.filter((node: any) => isDocument(node) || isComment(node))
-          : [];
-      },
-      enumerable: true,
-    },
+Object.defineProperty(YamlStream, 'content', {
```
char0n, line 10:
```suggestion
Object.defineProperty(YamlStream.prototype, 'content', {
```

char0n (approved):
LGTM! Nicely done! I'll follow up with some additional changes

PR #3690, opened 2024-01-17
refactor(ast): followup changes for YAML 1.2 Schemas

char0n:
Refs #3481


packages/apidom-ast/src/yaml/schemas/ScalarTag.ts
```diff
@@ -38,5 +38,6 @@ class ScalarTag {
     return node;
   }
 }
+/* eslint-enable class-methods-use-this */
```
char0n, line 41:
```suggestion
```

PR #3720, opened 2024-01-23
refactor(ns-workflows-1): change from stamps to TypeScript classes

glowcloud:
Refs #3481


char0n (commented):
Looks good! I've issues a PR that will serve as a CR. Please have a loot at it at https://github.com/swagger-api/apidom/pull/3726

PR #3791, opened 2024-02-05
chore(converter): use source code from the monorepo for dependencies

char0n:


packages/apidom-converter/package.json
char0n, line 1: When package is in private mode, all monorepo deps needs to be installed via `*`. When outside of private mode (released), latest version needs to be assigned manually, and from this moment lerna will take over the version bumps.

Issue #3788, opened 2024-02-05
parser-adapter-yaml-1-2: incorrect trimming


char0n:
The parser incorrectly parses empty single quote and double quote scalars.

```yaml
key: " "
```

```yaml
key: ' '
```

The above string, when parsed using `apidom-parser-adapter-yaml-1-2` returns ApiDOM that contains the empty string instead of one space string. I suspect issue comes from `canonical-format.ts` within the `apidom-ast` package. 

Specifically `formatFlowSingleQuoted` and `formatFlowDoubleQuoted` are affected.

char0n:
Addressed in https://github.com/swagger-api/apidom/pull/3792

PR #3792, opened 2024-02-05
fix(parser-adapter-yaml-1-2): incorrect quote scalars trimming

glowcloud:
Refs #3788


packages/apidom-parser-adapter-yaml-1-2/test/syntactic-analysis/indirect/visitors/CstVisitor.ts
char0n, line 1: Can you move all these tests into `test/adapter-nodes/index.ts` and `test/adapter-browser.ts` respectively?

Issue #3756, opened 2024-01-30
namespaces: refracting semantic ApiDOM


char0n:
Our namespaces are currently unable to refract already semantic ApiDOM. Simple change in code will allow **semantic** -> **semantic** conversion. This comes handy when there is a need to refract ApiDOM in OpenAPI 3.1.0 namespace to ApiDOM in OpenAPI 3.0.x namespace.

This issue has already been fixed for OpenAPI 3.0.x namespaces in https://github.com/swagger-api/apidom/pull/3755/files#diff-9b2aa67bd04bd47fd6036e71fc00721ee99677ac0ba38eae0201a228fb700f3eR84

Note all namespaces except OpenAPI 3.0.x are affected.

char0n:
Addressed in https://github.com/swagger-api/apidom/pull/3798

PR #3798, opened 2024-02-06
fix: refracting semantic ApiDOM

glowcloud:
Refs #3756


glowcloud:
This fix for AsyncAPI was already done here: #3786

char0n (approved):

PR #3802, opened 2024-02-06
feat(converter): add securitySchemeTypeRefractorPlugin

kowalczyk-krzysztof:
refs: SWG-9963

### DIFF
--- a/packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
+++ b/packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
@@ -1,0 +1,87 @@
+import {
+  ComponentsElement,
+  OpenApi3_1Element,
+  SecurityRequirementElement,
+  SecuritySchemeElement,
+  isSecuritySchemeElement,
+  isComponentsElement,
+} from '@swagger-api/apidom-ns-openapi-3-1';
+import { AnnotationElement, toValue, isObjectElement, Element } from '@swagger-api/apidom-core';
+
+import type { Toolbox } from '../toolbox';
+
+type SecuritySchemeTypePluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const securitySchemeTypeRefractorPlugin =
+  ({ annotations }: SecuritySchemeTypePluginOptions) =>
+  (toolbox: Toolbox) => {
+    const removedSecuritySchemes: SecuritySchemeElement[] = [];
+    const createAnnotation = <T extends Element>(element: T) =>
+      toolbox.createAnnotation.fromElement(
+        element,
+        'The "mutualTLS" type Security Scheme Object is not supported in OpenAPI 3.0.3. As a result, all Security Scheme Objects specified with the "mutualTLS" type have been removed.',
+        { classes: ['error'] },
+        { code: 'mutualTLS' },
+      );
+
+    return {
+      visitor: {
+        OpenApi3_1Element(element: OpenApi3_1Element) {
+          if (!isComponentsElement(element.components)) return undefined;
+          if (!isObjectElement(element.components.securitySchemes)) return undefined;
+
+          element.components.securitySchemes.forEach((value) => {
+            if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
+              removedSecuritySchemes.push(value);
+            }
+          });
+
+          return undefined;
+        },
+        ComponentsElement(element: ComponentsElement) {
+          if (!isObjectElement(element.securitySchemes)) return undefined;
+
+          element.securitySchemes.forEach((value, key) => {
+            if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
+              if (!removedSecuritySchemes.includes(value)) removedSecuritySchemes.push(value);
+              (element.securitySchemes as SecuritySchemeElement).remove(toValue(key));
+              annotations.push(createAnnotation(value));
+            }
+          });
+
+          return undefined;
+        },
+        SecurityRequirementElement(element: SecurityRequirementElement) {
+          if (!removedSecuritySchemes.length) return undefined;
+
+          const keysToRemove: string[] = [];
+
+          element.forEach((value, key) => {
+            const removedSecurityScheme = removedSecuritySchemes.find(
+              (securityScheme) => toValue(securityScheme.name) === toValue(key),
+            );
+
+            if (isSecuritySchemeElement(removedSecurityScheme)) {
+              keysToRemove.push(toValue(key));
+              annotations.push(createAnnotation(value));
+            }
+          });
+
+          if (!keysToRemove.length) return undefined;
+
+          keysToRemove.forEach((key) => {
+            element.remove(key);
+          });
+
+          return undefined;
+        },
+      },
+      post() {
+        removedSecuritySchemes.length = 0;
+      },
+    };
+  };
+
+export default securitySchemeTypeRefractorPlugin;
--- a/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/index.ts
+++ b/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/index.ts
@@ -1,0 +1,79 @@
+import path from 'node:path';
+import { mediaTypes as openAPI31MediaTypes } from '@swagger-api/apidom-parser-adapter-openapi-json-3-1';
+import { mediaTypes as openAPI30MediaTypes } from '@swagger-api/apidom-parser-adapter-openapi-json-3-0';
+import { AnnotationElement, includesClasses, toJSON, toValue } from '@swagger-api/apidom-core';
+import { assert, expect } from 'chai';
+
+import convert from '../../../../../src';
+
+describe('converter', function () {
+  context('strategies', function () {
+    context('openapi-3-1-to-openapi-3-0-3', function () {
+      context('security-scheme-type', function () {
+        specify(
+          'should remove SecurityScheme object if it has "mutualTLS" type',
+          async function () {
+            const fixturePath = path.join(__dirname, 'fixtures', 'security-scheme-type.json');
+            const convertedParseResult = await convert(fixturePath, {
+              convert: {
+                sourceMediaType: openAPI31MediaTypes.findBy('3.1.0', 'json'),
+                targetMediaType: openAPI30MediaTypes.findBy('3.0.3', 'json'),
+              },
+            });
+
+            expect(toJSON(convertedParseResult.api!, undefined, 2)).toMatchSnapshot();
+          },
+        );
+
+        specify('should create ERROR annotation', async function () {
+          const fixturePath = path.join(__dirname, 'fixtures', 'security-scheme-type.json');
+          const convertedParseResult = await convert(fixturePath, {
+            convert: {
+              sourceMediaType: openAPI31MediaTypes.findBy('3.1.0', 'json'),
+              targetMediaType: openAPI30MediaTypes.findBy('3.0.3', 'json'),
+            },
+          });
+          const annotations = Array.from(convertedParseResult.annotations);
+          const annotation = annotations.find((a: AnnotationElement) =>
+            a.code?.equals('mutualTLS'),
+          );
+
+          assert.isDefined(annotation);
+          assert.lengthOf(annotations, 2);
+          assert.isTrue(includesClasses(['error'], annotation));
+        });
+
+        specify('should attach source map to annotation', async function () {
+          const fixturePath = path.join(__dirname, 'fixtures', 'security-scheme-type.json');
+          const convertedParseResult = await convert(fixturePath, {
+            parse: {
+              parserOpts: { sourceMap: true },
+            },
+            convert: {
+              sourceMediaType: openAPI31MediaTypes.findBy('3.1.0', 'json'),
+              targetMediaType: openAPI30MediaTypes.findBy('3.0.3', 'json'),
+            },
+          });
+          const annotations = Array.from(convertedParseResult.annotations);
+          const annotation: AnnotationElement = annotations.find((a: AnnotationElement) =>
+            a.code?.equals('mutualTLS'),
+          );
+          const sourceMap = annotation.meta.get('sourceMap');
+          const { positionStart, positionEnd } = sourceMap;
+          const [startRow, startColumn, startChar] = toValue(positionStart);
+          const [endRow, endColumn, endChar] = toValue(positionEnd);
+
+          assert.isDefined(sourceMap);
+
+          assert.strictEqual(startRow, 9);
+          assert.strictEqual(startColumn, 32);
+          assert.strictEqual(startChar, 188);
+
+          assert.strictEqual(endRow, 12);
+          assert.strictEqual(endColumn, 13);
+          assert.strictEqual(endChar, 247);
+        });
+      });
+    });
+  });
+});
--- a/packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/index.ts
+++ b/packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/index.ts
@@ -18,6 +18,7 @@ import {
 import ConvertStrategy, { IFile } from '../ConvertStrategy';
 import openAPIVersionRefractorPlugin from './refractor-plugins/openapi-version';
 import webhooksRefractorPlugin from './refractor-plugins/webhooks';
+import securitySchemeTypeRefractorPlugin from './refractor-plugins/security-scheme-type';
 import type { ConverterOptions } from '../../options';
 import createToolbox from './toolbox';
 
@@ -56,7 +57,11 @@ class OpenAPI31ToOpenAPI30ConvertStrategy extends ConvertStrategy {
     const annotations: AnnotationElement[] = [];
     const parseResultElement = dispatchRefractorPlugins(
       cloneDeep(file.parseResult),
-      [openAPIVersionRefractorPlugin(), webhooksRefractorPlugin({ annotations })],
+      [
+        openAPIVersionRefractorPlugin(),
+        webhooksRefractorPlugin({ annotations }),
+        securitySchemeTypeRefractorPlugin({ annotations }),
+      ],
       {
         toolboxCreator: createToolbox,
         visitorOptions: { keyMap, nodeTypeGetter: getNodeType },
--- a/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/__snapshots__/index.ts.snap
+++ b/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/__snapshots__/index.ts.snap
@@ -1,0 +1,38 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`converter strategies openapi-3-1-to-openapi-3-0-3 security-scheme-type should remove SecurityScheme object if it has "mutualTLS" type 1`] = `
+{
+  "openapi": "3.0.3",
+  "paths": {
+    "/foo": {
+      "get": {
+        "summary": "foo",
+        "operationId": "foo",
+        "security": [
+          {},
+          {
+            "apiKey-scheme": [
+              "read",
+              "write"
+            ]
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "foo"
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "securitySchemes": {
+      "apiKey-scheme": {
+        "type": "apiKey",
+        "name": "apiKey-scheme",
+        "in": "header"
+      }
+    }
+  }
+}
+`;
--- a/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/fixtures/security-scheme-type.json
+++ b/packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/fixtures/security-scheme-type.json
@@ -1,0 +1,44 @@
+{
+  "openapi": "3.1.0",
+  "paths": {
+    "/foo": {
+      "get": {
+        "summary": "foo",
+        "operationId": "foo",
+        "security": [
+          {
+            "mutualTLS-scheme": [
+              "read",
+              "write"
+            ]
+          },
+          {
+            "apiKey-scheme": [
+              "read",
+              "write"
+            ]
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "foo"
+          }
+        }
+      }
+    }
+  },
+  "components": {
+    "securitySchemes": {
+      "mutualTLS-scheme": {
+        "type": "mutualTLS",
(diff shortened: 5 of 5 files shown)


packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,28 @@
+import { SecuritySchemeElement } from '@swagger-api/apidom-ns-openapi-3-1';
+import { AnnotationElement } from '@swagger-api/apidom-core';
```
char0n, line 2:
```suggestion
import { AnnotationElement, toValue } from '@swagger-api/apidom-core';
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,28 @@
...
+type SecuritySchemeTypePluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const securitySchemeTypeRefractorPlugin =
+  ({ annotations }: SecuritySchemeTypePluginOptions) =>
+  () => ({
+    visitor: {
+      SecuritySchemeElement(element: SecuritySchemeElement) {
+        if (element.get('type')?.toValue() === 'mutualTLS') {
```
char0n, line 13:
```suggestion
        if (toValue(element.get('type')) === 'mutualTLS') {
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,28 @@
...
+          const annotation = new AnnotationElement(
+            '"mutualTLS" type is not supported in OpenAPI 3.0.3.',
+            { classes: ['warning'] },
```
char0n, line 16: I would go for `error` level here. `mututalTLS` is not supported in OpenAPI 3.0.3. If we're just leaving the wrong value there (`mutualTLS`), then the tooling consuming the output of the converter will most probably fail. 

What is the overall strategy of handling these cases? 

```suggestion
            { classes: ['error'] },
```

char0n (changes requested):

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,28 @@
...
+            { code: 'mutualTLS' },
+          );
+
+          annotations.push(annotation);
+        }
+
+        return element;
```
char0n, line 23: If we don't want to do nothing with the element, we're just return `undefined`. More info in https://github.com/swagger-api/apidom/blob/main/packages/apidom-ast/src/traversal/visitor.ts#L144

```suggestion
        return undefined;
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
kowalczyk-krzysztof: The strategy is determined case by case. I just checked again the design document (see: SWG-9731) and indeed it seems I should be mutating the value here (`Annotate and remove this type/leave empty string`) so I will change the implementation.

packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/index.ts
```diff
@@ -0,0 +1,45 @@
...
+            convert: {
+              sourceMediaType: openAPI31MediaTypes.findBy('3.1.0', 'json'),
+              targetMediaType: openAPI30MediaTypes.findBy('3.0.3', 'json'),
+            },
+          });
+
+          expect(toJSON(convertedParseResult.api!, undefined, 2)).toMatchSnapshot();
+        });
+
+        specify('should create WARNING annotation', async function () {
```
char0n, line 25:
```suggestion
        specify('should create ERROR annotation', async function () {
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,27 @@
...
+        if (toValue(element.type) !== 'mutualTLS') return undefined;
+
+        const annotation = new AnnotationElement(
+          '"mutualTLS" type is not supported in OpenAPI 3.0.3. It will be removed from the converted document.',
+          { classes: ['error'] },
+          { code: 'mutualTLS' },
+        );
+
+        annotations.push(annotation);
+        return new SecuritySchemeElement();
```
char0n, line 22: `cloneShallow` comes from `@swagger-api/apidom-core`.

```suggestion
        const securitySchemeElementCopy = cloneShallow(element);
        securitySchemeElementCopy.remove('type');
        
        return securitySchemeElementCopy;
```

> Annotate and remove this type/leave empty string

I would assume this means that we want to remove the value of the `SecurityScheme.type` fixed field. But we're removing all fields from the `SecurityScheme` and basically creating and empty Security Scheme Object. If that was the intention then the code is correct. If the intention was to just remove the `type` fixed field, the code needs to look like I've drafter above.

char0n:
This is how we remove the the SecurityElement where ever it is inside the tree.
```js
const securitySchemeTypeRefractorPlugin =
  ({ annotations }: SecuritySchemeTypePluginOptions) =>
  () => ({
    visitor: {
      SecuritySchemeElement(element: SecuritySchemeElement) {
        if (toValue(element.type) !== 'mutualTLS') return undefined;

        const annotation = new AnnotationElement(
          '"mutualTLS" type is not supported in OpenAPI 3.0.3. It will be removed from the converted document.',
          { classes: ['error'] },
          { code: 'mutualTLS' },
        );
        annotations.push(annotation);

        return null; // remove ApiDOM Node from the Tree
      },
    },
  });

export default securitySchemeTypeRefractorPlugin;
```

kowalczyk-krzysztof:
As we discussed, the strategy is going to be different so I'm converting this PR to draft as I'm applying those changes.

kowalczyk-krzysztof:
@char0n I updated the solution but I'm not sure if I'm handling it correctly (checking for `name` match). Could use some more feedback.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,65 @@
...
+    const removedSecuritySchemes: SecuritySchemeElement[] = [];
+    const annotation = new AnnotationElement(
+      '"mutualTLS" type is not supported in OpenAPI 3.0.3. It will be removed from the converted document.',
+      { classes: ['error'] },
+      { code: 'mutualTLS' },
+    );
+
+    return {
+      visitor: {
+        OpenApi3_1Element(element: OpenApi3_1Element) {
```
char0n, line 25: Here is the version of `OpenApi3_1Element` I'm proposing:

```ts
import { isArrayElement } from '@swagger-api/apidom-core';
import { isComponentsElement, isSecuritySchemeElement } from '@swagger-api/apidom-ns-openapi-3-0/src';

OpenApi3_1Element(element: OpenApi3_1Element) {
  if (!isComponentsElement(element.components)) return undefined;
  if (!isArrayElement(element.components.securitySchemes)) return undefined;

  element.components.securitySchemes.forEach((item) => {
    if (isSecuritySchemeElement(item) && toValue(securityScheme.type) === 'mutualTLS') {
      removedSecuritySchemes.push(securityScheme);
    }
  });
}
```

Reasoning:

1. we cannot assume that `element?.components?.securitySchemes` is array. It can be a string or null. We parse the document as it is, we don't apply any validation. So we need to do assertion on everything.
2. again we cannot assume that `const securityScheme = el as SecuritySchemeElement;` is Security Scheme. It can be null, undefined, or anything else.

---

> NOTE: `isSecuritySchemeElement` was added in https://github.com/swagger-api/apidom/pull/3810

kowalczyk-krzysztof: Thanks for feedback. I added those missing checks. However it seems that `isObjectElement` should be used for `element.components.securitySchemes` as `securitySchemes` has type `ObjectElement | undefined` and `isArrayElement` for `securitySchemes` never returns `true`.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,85 @@
...
+  cloneShallow,
+  isObjectElement,
+} from '@swagger-api/apidom-core';
+import { isComponentsElement } from '@swagger-api/apidom-ns-openapi-3-0';
+
+type SecuritySchemeTypePluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const isSecuritySchemeElement = (element: SecuritySchemeElement) => !!element; // TODO: Add import from @swagger-api/apidom-ns-openapi-3-1 when new release is released
```
char0n, line 19: I've just added https://github.com/swagger-api/apidom/pull/3811, which adds `isSecuritySchemeElement` to the OpenAPI 3.1.0 namespace (which is what we really need here and not OpenAPI 3.0 predicate).

```js
import { isComponentsElement, isSecuritySchemeElement } from '@swagger-api/apidom-ns-openapi-3-1';
```

This will work. You don't need to wait for the release. Pull the latest changes from `main`, merge it to this branch, run `npm run build:cjs` && `npm run typescript:declaration` from the top of the monorepo, and the code in this PR will automatically start seeing the predicate.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,85 @@
...
+  SecurityRequirementElement,
+  SecuritySchemeElement,
+} from '@swagger-api/apidom-ns-openapi-3-1';
+import {
+  AnnotationElement,
+  toValue,
+  cloneShallow,
+  isObjectElement,
+} from '@swagger-api/apidom-core';
+import { isComponentsElement } from '@swagger-api/apidom-ns-openapi-3-0';
```
char0n, line 13:
```suggestion
import { isComponentsElement } from '@swagger-api/apidom-ns-openapi-3-1';
```

Our source is OpenAPI 3.1.0 and not OpenAPI 3.0.x.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
char0n: > has type ObjectElement | undefined

Yes, you're absolutely right, I did a mistake there. Of course it's an object.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,85 @@
...
+      { code: 'mutualTLS' },
+    );
+
+    return {
+      visitor: {
+        OpenApi3_1Element(element: OpenApi3_1Element) {
+          if (!isComponentsElement(element.components)) return undefined;
+          if (!isObjectElement(element.components.securitySchemes)) return undefined;
+
+          element.components.securitySchemes.forEach((securityScheme) => {
```
char0n, line 37:
```suggestion
          element.components.securitySchemes.forEach((value) => {
```

In ObjectElement context, the first argument here is a member `value` (second is key and third member), so it's better to call it that. In ArrayElement context the first argument is a current `item`.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
kowalczyk-krzysztof: Ah. Automatic import suggestion fails me once again. I should really double check those 😅

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
kowalczyk-krzysztof: Thanks, this worked. I was aware of the build step and i did that but indeed `npm run typescript:declaration` was needed too.

char0n:
We now have working `OpenApi3_1Element` visitor hook. I'm now looking at `ComponentsElement` hook.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,80 @@
...
+
+          element.components.securitySchemes.forEach((value) => {
+            if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
+              removedSecuritySchemes.push(value);
+            }
+          });
+
+          return undefined;
+        },
+        ComponentsElement(element: ComponentsElement) {
```
char0n, line 44: In light of https://github.com/swagger-api/apidom/pull/3812 here is my suggestion for `ComponentsElement`:

```js
ComponentsElement(element: ComponentsElement) {
  if (!isObjectElement(element.securitySchemes)) return undefined;

  element.securitySchemes.forEach((value, key) => {
    if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
      if (!removedSecuritySchemes.includes(value)) removedSecuritySchemes.push(value);
      element.remove(toValue(key));
    }
  });
}
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,80 @@
...
+          const clonedSecuritySchemes = cloneShallow(securitySchemes);
+
+          removedSecuritySchemes.forEach((securityScheme) => {
+            clonedSecuritySchemes.remove(toValue(securityScheme.name));
+            annotations.push(annotation);
+          });
+
+          return clonedSecuritySchemes;
+        },
+        SecurityRequirementElement(element: SecurityRequirementElement) {
```
char0n, line 57: Pseudocode for `SecurityRequirementElement`:

```ts
SecurityRequirementElement(element: SecurityRequirementElement) {
  if (removedSecuritySchemes.length === 0) return undefined;

  const keysToRemove: string[] = [];

  element.forEach((value, key) => {
    const removedSecurityScheme = removedSecuritySchemes.find((s) => toValue(s.name) === toValue(key))
    
    if (isSecuritySchemeElement(removedSecurityScheme)) {
      keysToRemove.push(toValue(key))
    }
  });

  keysToRemove.forEach((key) => { element.remove(key) });

  return undefined;
}
```

kowalczyk-krzysztof:
@char0n Two things I'm unsure about:

1. When exactly should I push annotations? Is the current implementation too much?
2. As it is, the mutated `SecurityRequirements` looks like the code below - there is an empty object left. Should something be done about it?
```json
"security": [
          {},
          {
            "apiKey-scheme": [
              "read",
              "write"
            ]
          }
        ],
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,85 @@
...
+              if (!removedSecuritySchemes.includes(value)) removedSecuritySchemes.push(value);
+              (element.securitySchemes as SecuritySchemeElement).remove(toValue(key));
+              annotations.push(annotation);
+            }
+          });
+
+          return undefined;
+        },
+        SecurityRequirementElement(element: SecurityRequirementElement) {
+          if (!removedSecuritySchemes.length || !isSecurityRequirementElement(element))
```
char0n, line 54:
```suggestion
          if (!removedSecuritySchemes.length)
```

It it guaranteed that when the `SecurityRequirementElement` hook executes it's only paramter will be of `SecurityRequirementElement` instance. No need for checks here.

char0n:
> 1. When exactly should I push annotations? Is the current implementation too much?

Good question. Let me think about this for a minute.

> 2. As it is, the mutated `SecurityRequirements` looks like the code below - there is an empty object left. Should something be done about it?
> 
> ```json
> "security": [
>           {},
>           {
>             "apiKey-scheme": [
>               "read",
>               "write"
>             ]
>           }
>         ],
> ```

We don't care. Empty Security Requirement Object is valid as empty object according to OpenAPI 3.0.x spec.

char0n:
> When exactly should I push annotations? Is the current implementation too much?

I guess this is dependent on the broader strategy how annotations should be generated. Do we want to generate annotation for every removed Schema Element or just generate one for the entire plugin?


I currently see two possible strategies for this. Let's stay in the realm of `Security Scheme Object` and model the strategies to that usecase.

### Generate single annotation for plugin run

We'll just introduce `semaphore` variable into the plugin and in `post` we'll add the annotation if semaphore is set to true. semaphore is set to true whenever the `Security Scheme Object` is removed from the definition.

Another option is to add annotation when first `Security Scheme Object` is removed and guard that it's not added later again. This has advantage over the above `post` solution in retaining the real order of annotations as they are added right when they are generated.

### Generate annotations for every element

Now this one would be more complex. We add annotation for every removal. Along with that we will translate element source maps to the annotation. This means that this single plugin can generate number of annotations and each annotation will contain  specific source map information about the removed Security Scheme Object.

char0n:
> I think users of the converter would prefer to see information about all changes to the API definition and where they happened. So IMO Generate annotations for every element would work best.

All right, decided then.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
```diff
@@ -0,0 +1,83 @@
...
+type SecuritySchemeTypePluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const securitySchemeTypeRefractorPlugin =
+  ({ annotations }: SecuritySchemeTypePluginOptions) =>
+  () => {
+    const removedSecuritySchemes: SecuritySchemeElement[] = [];
+    const annotation = new AnnotationElement(
+      '"mutualTLS" type is not supported in OpenAPI 3.0.3. It will be removed from the converted document.',
```
char0n, line 20:
```suggestion
      'The "mutualTLS" type Security Scheme Object is not supported in OpenAPI 3.0.3. As a result, all Security Scheme Objects specified with the "mutualTLS" type have been removed.',
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type.ts
char0n, line 1: THis is how we can use annotation toolbox utils now that we merged https://github.com/swagger-api/apidom/pull/3816:

```ts
import { Element, AnnotationElement } from '@swagger-api/apidom-core';
import type { Toolbox } from '../toolbox';

const securitySchemeTypeRefractorPlugin =
  ({ annotations }: SecuritySchemeTypePluginOptions) =>
  (toolbox: Toolbox) => {
    const removedSecuritySchemes: SecuritySchemeElement[] = [];
    const createAnnotation = <T extends Element>(element: T) => toolbox.createAnnotation.fromElement(
        element,
        'The "mutualTLS" type Security Scheme Object is not supported in OpenAPI 3.0.3. As a result, all Security Scheme Objects specified with the "mutualTLS" type have been removed.',
        { classes: ['error'] },
        { code: 'mutualTLS' },
    );

    return {
      visitor: {
        OpenApi3_1Element(element: OpenApi3_1Element) {
          if (!isComponentsElement(element.components)) return undefined;
          if (!isObjectElement(element.components.securitySchemes)) return undefined;

          element.components.securitySchemes.forEach((value) => {
            if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
              removedSecuritySchemes.push(value);
            }
          });

          return undefined;
        },
        ComponentsElement(element: ComponentsElement) {
          if (!isObjectElement(element.securitySchemes)) return undefined;

          element.securitySchemes.forEach((value, key, member) => {
            if (isSecuritySchemeElement(value) && toValue(value.type) === 'mutualTLS') {
              if (!removedSecuritySchemes.includes(value)) removedSecuritySchemes.push(value);
              (element.securitySchemes as SecuritySchemeElement).remove(toValue(key));
              annotations.push(createAnnotation(member));
            }
          });

          return undefined;
        },
        SecurityRequirementElement(element: SecurityRequirementElement) {
          if (!removedSecuritySchemes.length) return undefined;

          const keysToRemove: string[] = [];

          element.forEach((value, key, member) => {
            const removedSecurityScheme = removedSecuritySchemes.find(
              (securityScheme) => toValue(securityScheme.name) === toValue(key),
            );

            if (isSecuritySchemeElement(removedSecurityScheme)) {
              keysToRemove.push(toValue(key));
              annotations.push(createAnnotation(member);
            }
          });

          if (!keysToRemove.length) return undefined;

          keysToRemove.forEach((key) => element.remove(key));

          return undefined;
        },
      },
      post() {
        removedSecuritySchemes.length = 0;
      },
    };
  };
```

packages/apidom-converter/test/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/security-scheme-type/index.ts
```diff
@@ -0,0 +1,48 @@
...
+            a.code?.equals('mutualTLS'),
+          );
+
+          assert.isDefined(annotation);
+          assert.lengthOf(annotations, 2);
+          assert.isTrue(includesClasses(['error'], annotation));
+        });
+      });
+    });
+  });
```
char0n, line 47: We need additional test for attaching source maps to annotations. Pseudo code shows how to achieve that:

```js
specify('should attach source maps annotation', async function () {
  const fixturePath = path.join(__dirname, 'fixtures', 'security-scheme-type.json');
  const convertedParseResult = await convert(fixturePath, {
    parse: {
      parserOpts: { sourceMap: true },
    },
    convert: {
      sourceMediaType: openAPI31MediaTypes.findBy('3.1.0', 'json'),
      targetMediaType: openAPI30MediaTypes.findBy('3.0.3', 'json'),
    },
  });
  const annotations = Array.from(convertedParseResult.annotations);
  const annotation = annotations.find((a: AnnotationElement) => a.code?.equals('mutualTLS'));
  
  toValue(annotation.meta.get('sourceMap')); // assertions  
});

```

char0n (approved):
LGTM

Issue #3697, opened 2024-01-18
ApiDOM converter


char0n:
### OpenAPI Changelog

I would recommend [Mike Ralphson’s article](https://dev.to/mikeralphson/openapi-31-the-gnarly-bits-58d0) describing OpenAPI 3.1.0 changes comparing to OpenAPI 3.0.x, which I found to be best source of truth related to changes between the relevant versions. It’s even better than [official OpenAPI 3.1.0 release log](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0). Another source of possibly valuable info: https://github.com/apiture/openapi-down-convert. The spotlight article also contains wrong information about JSON Schema version used in OpenAPI 3.1.0. The version used in OpenAPI 3.1.0 is actually JSON Schema Draft 2020-12 and not 2019-09.  

### Target versions

We're targeting OpenAPI 3.0.3 (latest released version of OpenAPI 3.0 release branch). OpenAPI 3.0.4[ is currently in making](https://github.com/OAI/OpenAPI-Specification/issues/3176).

### Doing conversion on resolved definition

To clarify terms a bit for future reference: “Resolved” wording is used in original Swagger JavaScript tooling to refer to dereferencing. Dereferencing is part of the nomenclature used by JSON Schema and ApiDOM and refers to the mechanism of reference removal. In JSON Schema and ApiDOM context “resolve” refers more to how URIs are resolved and retrieved.

For converted to work the dereferenced OpenAPI 3.1.0 would have to eliminate possible cycles so that the dereferenced OpenAPI 3.1.0 can be easily traversed and converted to OpenAPI 3.0.3 and serialized in YAML 1.2 or JSON. Since JSON Schema 2019-09 the[ reference removal is no longer safe or possible](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.appendix.B.2). Which means ApiDOM dereferencing mechanism is just an approximation of the ideal state.

[...]

PR #3848, opened 2024-02-22
feat(converter): add infoSummaryRefractorPlugin

kowalczyk-krzysztof:
This plugin removes the `summary` fixed field of `Info Object`.

Refs SWG-9954

PR #3885, opened 2024-03-01
feat(converter): add Reference Object summary and description plugins

tekyu:
his plugin removes the `summary` and `description` fields of `Reference Object`.

Refs SWG-10235
Refs #3697


packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/reference-summary-description.ts
```diff
@@ -0,0 +1,39 @@
...
+        ReferenceElement(element: ReferenceElement) {
+          if (element.hasKey('summary')) {
+            annotations.push(summaryAnnotation);
+            element.remove('summary');
+          }
+          if (element.hasKey('description')) {
+            annotations.push(descriptionAnnotation);
+            element.remove('description');
+          }
+          return undefined;
```
char0n, line 33: No need to return here. the `return` will be implicit and equivalent if we remove this code.

```suggestion
          return undefined;
```

char0n:
@tekyu there is an option to have this plugin divided into two. One plugin that removes summary, one that removes description. The idea is that every plugin just handles one aspect of the the transformation.

In this particular case, we can say that the aspect is to remove the non-supported fields, or we can divide the aspect into two separate ones. 

Please let me know what you think.

tekyu:
@char0n if the idea is to have plugins that handles single operation I don't want to disrupt that with my plugin. I'll split them then

char0n:
@tekyu it's your decision, as for this particular one I don't have strong opinion.

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/reference-description.ts
```diff
@@ -0,0 +1,29 @@
...
+import { AnnotationElement } from '@swagger-api/apidom-core';
+
+type ReferenceDescriptionPluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const referenceDescriptionRefractorPlugin =
+  ({ annotations }: ReferenceDescriptionPluginOptions) =>
+  () => {
+    const descriptionAnnotation = new AnnotationElement(
```
char0n, line 11:
```suggestion
    const annotation = new AnnotationElement(
```

packages/apidom-converter/src/strategies/openapi-3-1-to-openapi-3-0-3/refractor-plugins/reference-summary.ts
```diff
@@ -0,0 +1,29 @@
...
+import { AnnotationElement } from '@swagger-api/apidom-core';
+
+type ReferenceSummaryPluginOptions = {
+  annotations: AnnotationElement[];
+};
+
+const referenceSummaryRefractorPlugin =
+  ({ annotations }: ReferenceSummaryPluginOptions) =>
+  () => {
+    const summaryAnnotation = new AnnotationElement(
```
char0n, line 11:
```suggestion
    const annotation = new AnnotationElement(
```

char0n (approved):
Two nitpicks related to naming, otherwise LGTM

char0n:
@tekyu plese use `Squash + merge` github feature described in https://github.com/swagger-api/apidom/pull/3848#issuecomment-1959582561


Please also use commit message format described in: https://github.com/swagger-api/apidom/pull/3848#issuecomment-1959505764

PR #4055, opened 2024-04-25
refactor(json-schema-draft-4): change from stamps to classes

glowcloud:
Refs #3481


packages/apidom-ns-asyncapi-2/src/refractor/visitors/async-api-2/schema/AllOfVisitor.ts
```diff
@@ -1,25 +1,28 @@
...
+  specificationObj as JSONSchemaDraft7Specification,
+  AllOfVisitorOptions,
+} from '@swagger-api/apidom-ns-json-schema-draft-7';
 
 import ReferenceElement from '../../../../elements/Reference';
 import { isReferenceElement } from '../../../../predicates';
 
+export type { AllOfVisitorOptions };
+
 const { allOf: JSONSchemaAllOfVisitor } =
```
glowcloud, line 12: I think we could do a simple import without `specificationObj` if we export AllOfVisitor in the corresponding `json-schema-draft` package like this for `Visitor`:
```js
export { default as Visitor } from './refractor/visitors/Visitor';
export type { VisitorOptions } from './refractor/visitors/Visitor';
```
but not sure about it, so left it like this for now.

char0n: The idea was that the `specificationObj` is a public API to get any visitor you need from the upstream package. 

We are only doing the following (exporting types) to avoid serializing types from `specificationObj` which can end up in crashing the `tsc` (if you remember).

```ts
export type { VisitorOptions } from './refractor/visitors/Visitor'; 
```

We don't really want to export every visitor as it would mean, every time you write a visitor you have to remember to export it and thus extend the npm package public API. We only want to explicitly export visitors which are used to build another more specialized visitors (SpecificationVisitor, MapVisitor, etc...)

In my mind by currently only exporting the types of all the visitors to avoid `tsc` crash and if we forget to export one, nothing serious happens, as it's just a typescript and not JavaScript.

char0n (approved):
Nice work!

Issue #4134, opened 2024-05-23
ns-openapi-3-1: allow normalization refractor plugins to work on scoped ApiDOM trees


char0n:
## Prelude

We have an [issue](https://github.com/swagger-api/swagger-js/issues/1396) with normalization during partial dereference in swagger-client. How it works currently is that the undereferenced definition gets normalized and then partially dereferenced. `$$normalized` flag is set on the dereferenced structure. This means that whatever got dereferencedd during dereferencing phase, will not be normalized. This creates various side effects and inconsistencies in derefencing. End use of SwaggerUI or SwaggerClient must not expected that anything that is referenced will get normalized.

More info in https://github.com/swagger-api/swagger-js/issues/1396#issuecomment-2126741753

---

### Optimization

Implementation aligned above will require additional traversal pass. The moment all the plugins are idempotent, we can try incorporating them into dereference traversal.

char0n:
All normalize refractor plugins have been made idempotent.

PR #4151, opened 2024-05-28
feat(ns-openapi-3-1): make operation ids refractor plugin idempotent

glowcloud:
Refs #4134


packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-operation-ids.ts
```diff
@@ -116,10 +133,27 @@ const plugin =
           },
         },
         OperationElement: {
-          enter(operationElement: OperationElement) {
+          enter(
+            operationElement: OperationElement,
+            key: string | number,
+            parent: Element | undefined,
+            elementPath: (string | number)[],
```
glowcloud, line 140: Not sure what to name this, considering that we get `path` from `paths` in this one:
```js
const path = last(paths) as string;
````

char0n: I'd rename the top level paths to pathTemplates and then use paths in visitor method arguments

char0n (approved):

Issue #4229, opened 2024-07-08
ns-openapi-3-1: normalize-parameter-examples refractor plugin incorrectly sets schema example as an array of examples 


glowcloud:
When normalizing a parameter such as
 ```json
      "name": {
        "name": "name",
        "in": "query",
        "description": "name of the pet",
        "required": false,
        "schema": {
          "type": "string",
          "example": "schema example"
        },
        "examples": {
          "example1": {
            "value": "examples example"
          }
        }
      }
```
the `example` field in the `schema` will be normalized to:
```js
"example": ['examples example']
```
Instead, it should be normalized to:
```js
"example": "examples example"
```

This happens because we set example to the array of examples:
https://github.com/swagger-api/apidom/blob/1a6da7f78fb90464e0c4c6ffabc38d9f3a6ac6a3/packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-parameter-examples.ts#L100-L103
instead, we should be setting it to the first example of the examples array like we do in the `normalize-header-examples` plugin:
https://github.com/swagger-api/apidom/blob/1a6da7f78fb90464e0c4c6ffabc38d9f3a6ac6a3/packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-header-examples/index.ts#L100-L103

char0n:
Addressed in https://github.com/swagger-api/apidom/pull/4230

PR #4230, opened 2024-07-08
fix(ns-openapi-3-1): set example to first element of examples in parameter examples refractor plugin

glowcloud:
Refs #4229


char0n (approved):
Good catch!

Issue #4298, opened 2024-08-08
apidom-reference: parser plugins are inflexible when it comes to file extensions


char0n:
### Current behavior

Parser plugins support `fileExtensions`. 

When this option is set to empty list, it means `fileExtensions` are not required and if provided, it's not checked what the extensions are.

When `fileExtensions` as set to e.g. `[.yaml, .yml]`, is means that these are the allowed and one of those value is required. If extension is omitted the parser rejects to parse.

This feels inflexible and by default doesn't support [SwaggerHUB Domains](https://support.smartbear.com/swaggerhub/docs/en/domains.html). 

### Proposed behavior

By default, `fileExtensions` should be empty list. Which means that the parser plugin will always run it's detection mechanism to determine the format and specification within the file object. No file or URL will be  rejected just because it doesn't have a proper file extension. 

Anybody that is looking for a strict behavior of requiring file extensions, it will still be possible to configure it, by just setting e.g. `[.yaml, .yml]` as `fileExtensions`.

char0n:
Addressed in https://github.com/swagger-api/apidom/pull/4299

PR #4299, opened 2024-08-08
feat(reference): loosen requirement for file extensions in parser plugins

char0n:
Refs #4298


glowcloud (approved):
Looks good, apart from the ESLint errors

Issue #4385, opened 2024-10-19
Switch `moduleResolution` in tsconfig to `nodenext`


char0n:
This will allow TypeScript to recognize `exports` field in npm packages. Without this setting on we're not able to use proper imports of sub-paths with appropriate typings. `nodenext` this setting will also require to have an explicit file extensions, which will require a lot of changes to a lot of files.

### TODO

This list is for the case when all packages has been transformed.

- [x] Global `tsconfig.json` needs to be modified when all packages are transformed
- [x] Local `tsconfig.json` needs to be amended to original form
- [x] Local `.eslintrc` needs to be dropped and made global
- [x] [Convert packages tests to nodenext as well](https://github.com/swagger-api/apidom/pull/4446#issuecomment-2461791531)

char0n:
In https://github.com/swagger-api/apidom/pull/4402, I've figured out how to migrate progressively monorepo package by package and transformed the `apidom-error` package.

char0n:
This ticket has been completed. Some packages run their tests still in legacy mode, due to the dependencies they are using.

PR #4446, opened 2024-11-05
build(apidom-core): use nodenext for TypeScript modules

glowcloud:
Refs #4385

### DIFF
--- a/packages/apidom-core/check-types-test.tsconfig.json
+++ b/packages/apidom-core/check-types-test.tsconfig.json
@@ -1,12 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "module": "nodenext",
-    "moduleResolution": "nodenext",
     "allowImportingTsExtensions": true
   },
   "include": [
-    "src/**/*",
-    "../@types/**/*.d.ts"
+    "test/**/*"
   ]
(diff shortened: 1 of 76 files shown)


char0n:
There is a problem checking types of the tests, because of the the vendor libraries that we use (e.g. mocha-chai-jest-snapshot) don't have proper ESM vs CJS module handling.

Another issue is that some of our own apidom packages used within tests are still  not converted to `nodenext` (the conversion is progressive). 

So we'll use separate typescript check for the source code and for the test. The source code typescript check will work in `nodenext` mode, but the test will work in legacy (current) global mode.

packages/apidom-core/check-types-test.tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "allowImportingTsExtensions": true
```
char0n, line 4: Does this need to be here? As we're stipulated, supporting `.js` extension is enabled by default

char0n: We're using '.ts` and not .`js` imports. So this is fine.

char0n (approved):
LGTM

PR #4495, opened 2024-11-14
build(ls): use nodenext for TypeScript modules

glowcloud:
Refs #4385

**NOTE:** had to add a few `ts-ignore` annotations to: [ajv-utils](https://github.com/swagger-api/apidom/blob/3b5e468e95b4b030d80de43f1109f16ebf0aec54/packages/apidom-ls/src/services/validation/providers/ajv-utils.ts) due to errors:
```
error TS2349: This expression is not callable.
  Type 'typeof import(".../node_modules/ajv-errors/dist/index")' has no call signatures.
```

```
error TS2351: This expression is not constructable.
  Type 'typeof import(".../node_modules/ajv/dist/ajv")' has no construct signatures.
```
I don't think we can fix it differently due to the library usage.


char0n (approved):
I will just trust you and CI as just openining this frezez my browser

PR #4429, opened 2024-10-30
feat(parser-adapter-yaml-1-2): use new @tree-sitter-grammars/tree-sitter-yaml grammar

char0n:
Refs #4033

PR #4502, opened 2024-11-17
fix(apidom-ls): fix YAML completion in end of doc scenario

frantuma:
This fixes wrong/absent completion items being returned in cases where YAML is not correct (e.g. partial key) and cursor is at beginning of line right after a non empty node (possibly related to the YAML tree-sitter grammar switch in #4429)


char0n (approved):
LGTM. Nice work!

Issue #4382, opened 2024-10-19
Switch rollup + rollup-plugin-dts with api-extractor.com


char0n:
[rollup-plugin-dts](https://github.com/Swatinem/rollup-plugin-dts) is in maintenance mode only. There is a risk that author will stop maintaining it and adding support for future TypeScript versions. `api-extractor` is a recommended MicroSoft tool for now for TypeScript bundling.

### TODO

- [x] manually test apidom-reference exports entrypoints and fix if needed

### Resources
- https://github.com/Swatinem/rollup-plugin-dts
- https://api-extractor.com/

char0n:
In https://github.com/swagger-api/apidom/pull/4422, I've figured out how to migrate progressively monorepo package by package and transformed the apidom-error package.

PR #4504, opened 2024-11-18
chore(ast): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382


packages/apidom-ast/src/index.ts
```diff
@@ -3,8 +3,10 @@ export { default as JsonNode } from './json/nodes/JsonNode.ts';
 export { default as JsonDocument } from './json/nodes/JsonDocument.ts';
 export { default as JsonObject } from './json/nodes/JsonObject.ts';
 export { default as JsonProperty } from './json/nodes/JsonProperty.ts';
+export type { JsonValue as JsonValueType } from './json/nodes/JsonProperty.ts';
```
glowcloud, line 6: This needs to have a different name as we're already export `JsonValue` class.
Perhaps `JsonPropertyValue` would be better?
```suggestion
export type { JsonValue as JsonPropertyValue } from './json/nodes/JsonProperty.ts';
```

char0n (approved):
Looks solid!

packages/apidom-ast/src/index.ts
char0n: Yes, go ahead with this.

PR #4514, opened 2024-11-20
chore(parser-adapter-yaml-1-2): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382


char0n (approved):
Approved, assuming test was done on node adapter api extraction?

glowcloud:
> Approved, assuming test was done on node adapter api extraction?

Yes, I checked with `adapter-node` entrypoint and there were no warnings.

PR #4516, opened 2024-11-20
chore(parser-adapter-json): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382


char0n (approved):
Approving, assuming test was done on node adapter api extraction.

PR #4532, opened 2024-11-20
chore(ns-json-schema-draft-4): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382

**IMPORTANT**: all of mixins had to be transformed from e.g.:
```js
class AllOfVisitor extends Mixin(SpecificationVisitor, ParentSchemaAwareVisitor, FallbackVisitor) {...}
```
to:
```js
export const AllOfVisitorBase = Mixin(
  SpecificationVisitor,
  ParentSchemaAwareVisitor,
  FallbackVisitor,
);

class AllOfVisitor extends AllOfVisitorBase {...}
```
as the type of mixin had to be exported due to a warning:
```
Warning: types/refractor/visitors/json-schema/AllOfVisitor.d.ts:18:1 - (ae-forgotten-export) The symbol "AllOfVisitor_base" needs to be exported by the entry point index.d.ts
```

Let me know if this is fine, and if so, if the names should have a different convention than `...Base`.


char0n:
We don't want to expose base as public API, it's something that should remain hidden. Let's filter and hide all the warnings containing `Visitor_base` substrings.

char0n (changes requested):
https://github.com/swagger-api/apidom/pull/4532#issuecomment-2490366287

char0n (approved):

char0n:
NOTE: shx was used to provide multiplatform `grep`

PR #4538, opened 2024-11-21
chore(reference): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382


packages/apidom-reference/src/index.ts
```diff
@@ -31,6 +32,13 @@ export type { BundleStrategyOptions } from './bundle/strategies/BundleStrategy.t
 
 export { default as options } from './options/index.ts';
 export { merge as mergeOptions } from './options/util.ts';
+export type {
+  ReferenceOptions as ApiDOMReferenceOptions,
```
glowcloud, line 36: Went with the `ApiDOM` prefix for these exports as `ReferenceOptions` has to be renamed due to another export:
```ts
export type { ReferenceOptions } from './Reference.ts';
```
Not sure what else we could go for, except for perhaps dropping the `Reference` part from all of them
```suggestion
  ReferenceOptions as Options,
  ReferenceBundleOptions as BundleOptions,
  ...
```

char0n (approved):

PR #4552, opened 2024-11-22
chore(ls): use API Extractor for TypeScript rollup

glowcloud:
Refs #4382


char0n (approved):

char0n:
Let's manually tests all the exports entry points. No automates needed at this point

Issue #4539, opened 2024-11-21
Can't resolve '@swagger-api/apidom-json-pointer' in latest release


benjdlambert:
Looks like the latest release that went out appears to have broken our tests. 

```
Error: Failed to compile '../../node_modules/@swagger-api/apidom-ns-openapi-3-1/src/refractor/toolbox.mjs':
  Module not found: Error: Can't resolve '@swagger-api/apidom-json-pointer' in '/tmp/backstage-e2e-wwR5QI/test-app/node_modules/@swagger-api/apidom-ns-openapi-3-1/src/refractor'
```

Wondering if this is a bad publish?

Additional build logs here: https://github.com/backstage/backstage/actions/runs/11952606327/job/33318690026

IagorSs:
Same error here

paztis:
swagger-client webpack bundling totally fails also for us

benjdlambert:
Looks like adding a resolution for `@swagger-api/apidom-json-pointer` works for us, but not ideal.

```diff
   "resolutions": {
     "@types/react": "^18",
     "@types/react-dom": "^18"
     "@types/react-dom": "^18",
+    "@swagger-api/apidom-json-pointer": "1.0.0-alpha.10"
```

paztis:
in fact it may be the same for all you optional dependencies.
```
  "optionalDependencies": {
    "@swagger-api/apidom-error": "^1.0.0-alpha.1",
    "@swagger-api/apidom-json-pointer": "^1.0.0-alpha.1",
    "@swagger-api/apidom-ns-asyncapi-2": "^1.0.0-alpha.1",
    "@swagger-api/apidom-ns-openapi-2": "^1.0.0-alpha.1",
    "@swagger-api/apidom-ns-openapi-3-0": "^1.0.0-alpha.1",
    "@swagger-api/apidom-ns-openapi-3-1": "^1.0.0-alpha.1",
    "@swagger-api/apidom-ns-workflows-1": "^1.0.0-alpha.1",
    "@swagger-api/apidom-parser-adapter-api-design-systems-json": "^1.0.0-alpha.1",
    "@swagger-api/apidom-parser-adapter-api-design-systems-yaml": "^1.0.0-alpha.1",
    "@swagger-api/apidom-parser-adapter-asyncapi-json-2": "^1.0.0-alpha.1",
    "@swagger-api/apidom-parser-adapter-asyncapi-yaml-2": "^1.0.0-alpha.1",
[...]

PR #4556, opened 2024-11-25
fix(reference): use upper bound <1.0.0-rc.0 for all optional deps

char0n:
Refs #4539


char0n:
I remember that before the learn will automatically remove the upper boundary. There is no way to keep it there. This is the last attempt

Issue #4384, opened 2024-10-19
playground: migrate build infra from CreateReactApp (CRA) to Vite


char0n:
The CRA deps are getting very old and insecure. CRA is EOL, and has been deprecated since early 2023.

glowcloud:
Addressed in https://github.com/swagger-api/apidom/pull/4570

PR #4570, opened 2024-11-29
build(playground): migrate from Create React App to Vite

glowcloud:
Refs #4384

For the removed files in `scripts`, it seems that the polyfills for `fs` and `path` which were disabled by this config:
```
  config.resolve.fallback = {
    ...config.resolve.fallback,
    fs: false,
    path: false,
  };
```

should be automatically disabled by Vite. When building the project, I get this warning:
```
[plugin vite:resolve] Module "fs" has been externalized for browser compatibility
```
which should mean that the polyfills are disabled: https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility


packages/apidom-playground/vite.config.mjs
```diff
@@ -0,0 +1,26 @@
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+
+export default defineConfig({
+  plugins: [
+    react(),
+    {
```
glowcloud, line 7: This plugin is a replacement for this part of the previous `start` script:
```
cross-env GENERATE_SOURCEMAP=false
``` 
Using other sourcemap configs did not disable them for JS files. 

Source: https://github.com/vitejs/vite/issues/9825#issuecomment-2413567622

packages/apidom-playground/vite.config.mjs
```diff
@@ -0,0 +1,28 @@
...
+  },
+  assetsInclude: ['**/*.wasm'],
+  build: {
+    sourcemap: false,
+    sourcemapExcludeSources: true,
+    target: '',
+    outDir: 'build',
+    chunkSizeWarningLimit: 1000,
+    rollupOptions: {
+      external: ['fs', 'path'],
```
char0n, line 19:
```suggestion
      output: { format: 'umd' },
      external: ['fs', 'path'],      
```

Issue #1819, opened 2022-08-09
JSON Schema Draft 2019-09 namespace 


char0n:
Scope of this issue is to implement JSON Schema Draft 2019-09 as an ApiDOM namespace.

glowcloud:
JSON Schema 2019-09 namespace was implemented in https://github.com/swagger-api/apidom/pull/4662.

PR #4658, opened 2025-01-16
feat(ns-json-schema-draft-6): handle inheritance of $schema and $id

glowcloud:
Refs #1819


packages/apidom-ns-json-schema-draft-6/src/refractor/visitors/json-schema/index.ts
```diff
@@ -43,6 +66,39 @@ class JSONSchemaVisitor extends Mixin(FixedFieldsVisitor, FallbackVisitor) {
 
     return result;
   }
+
+  handle$schema(objectElement: ObjectElement): void {
+    // handle $schema keyword in embedded resources
+    if (isUndefined(this.parent) && !isStringElement(objectElement.get('$schema'))) {
+      // no parent available and no $schema is defined, set default $schema
+      this.element.setMetaProperty('inherited$schema', this.default$schema);
+    } else if (isJSONSchemaElement(this.parent) && !isStringElement(objectElement.get('$schema'))) {
```
glowcloud, line 75: According to [Draft 6](https://json-schema.org/draft-06/draft-wright-json-schema-01#rfc.section.7)
```
The "$schema" keyword SHOULD be used in a root schema. It MUST NOT appear in subschemas.
```
Then if the `$schema` exists in a subschema, should we always override it here and set the `inherited$schema` from the parent?

char0n (approved):

PR #4870, opened 2025-04-30
refactor(json-pointer): use RFC compliant impl under the hood

char0n:
NOTE: this is a community contribution

---

The overall goal of this effort is to deprecate the apidom-json-pointer package, as it's not RFC compliant, contains many bugs and only works for ApiDOM.

---

I've developed fully compliant RCF 6901 implementation in https://github.com/swaggerexpert/json-pointer. The library works for JSON, Minim, ApiDOM, Map/Set and Immutable.js realms; which means we can use it with basic JSON in SwaggerClient, Immutable.js structures in SwaggerUI and Minim/ApiDOM element in ApiDOM. Parsing of JSON Pointer is based on real parser which allows to generate proper error messages with [inferred expectations](https://github.com/swaggerexpert/json-pointer?tab=readme-ov-file#tracing). Generated error messages are carefully constructed in a way to contain maximum possible information for user to understand what went wrong; errors are also decorated with fields containing additional information.  Evaluation can use diagnostics and tracings allowing us to detect what went wrong and where. [URI Fragment Identifier representation](https://github.com/swaggerexpert/json-pointer?tab=readme-ov-file#uri-fragment-identifier) is fully compliant with [RFC 3986, section 3.5](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5 ).

---

When/If this gets accepted/merged, I'll continue with the effort of replacing `apidom-json-pointer` package with https://github.com/swaggerexpert/json-pointer in ApiDOM and SwaggerClient.


packages/apidom-ns-openapi-3-1/src/refractor/toolbox.ts
```diff
@@ -68,6 +68,8 @@ export const ancestorLineageToJSONPointer = <T extends (Element | Element[])[]>(
     return path;
   }, [] as string[]);
 
+  console.dir(compileJSONPointerTokens(jsonPointerTokens));
```
glowcloud, line 71: Leftover debugging message.

glowcloud (approved):

packages/apidom-ns-openapi-3-1/src/refractor/toolbox.ts
char0n: Resolved, thanks!

char0n:
@glowcloud all green here. Ready for merge

PR #4940, opened 2025-05-30
fix(core): fix custom deep merge not affecting nested properties

glowcloud:


packages/apidom-core/test/merge/deepmerge.ts
```diff
@@ -455,6 +456,42 @@ describe('deepmerge', function () {
...
+      });
+      const merged = deepmerge(x, y, {
+        customMerge: (keyElement) => {
+          if (toValue(keyElement) === 'enum') {
+            return (targetElement, sourceElement) =>
+              new ArrayElement(
+                uniqWith(equals)([...toValue(targetElement), ...toValue(sourceElement)]),
+              );
+          }
+          return deepmerge;
```
glowcloud, line 481: According to the type of `CustomMerge` we should be returning `DeepMerge`, but perhaps it should be `DeepMerge | undefined`? Considering [this return](https://github.com/swagger-api/apidom/blob/5ac48c229ad46e927a903ad63956f5269c170920/packages/apidom-core/src/merge/deepmerge.ts#L118) from `getMergeFunction`, as well as the [example](https://github.com/swagger-api/apidom/blob/main/packages/apidom-core/README.md#custommerge) from the documentation: 
```ts
const mergeNames = (nameA: ObjectElement, nameB: ObjectElement) =>
  new StringElement(`${toValue(nameA.get('first'))} and ${toValue(nameB.get('first'))}`);
const customMerge = (key: Element) => (toValue(key) === 'name' ? mergeNames : undefined);
```

char0n (approved):

PR #4952, opened 2025-06-12
feat: add PoC of handling discriminator for oneOf/anyOf cases in OpenAPI 3.1

glowcloud:
Notes on the idea of handling discriminator:

**Dereference**: dereference value of `MemberElement` contained within `Discriminator.mapping` and set the result as its metadata.
```yaml
discriminator:
   propertyName: petType
   mapping:
      dog: Dog # => should be dereferenced and the dereference result should be set as metadata
```

**Normalization**: compare `oneOf/anyOf` items with the mapping values metadata and create normalized mapping based on the result:
- If both `oneOf` and `anyOf` are present, do not normalize
- Compare external references and internal references that do not point to `components/schema/<schemaName>`:
  - These references will be recognized by having a different `baseURI` or by not having `schemaName` metadata.
  - The comparison is done based on `$refBaseURI` metadata.
  - If no mapping exists, normalization will not be successful.
- Compare internal references that point to `components/schema/<schemaName>`:
  - These references will be recognized by having `schemaName` metadata.
  - The comparison is done based on `schemaName` and `$refBaseURI` metadata.
  - If no mapping exists, create it  as `<schemaName>: <schema>`.
- If normalization was successful, set the normalized mapping to a Specification Extension field (e.g. `x-normalized-mapping`)

Additionally, we should have a new strategy option to check if we should dereference Discriminator.mapping values.


packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  const metaRefFields = toValue(item.getMetaProperty('ref-fields'));
+  const metaRefOrigin = toValue(item.getMetaProperty('ref-origin'));
+  const metaSchemaName = toValue(item.getMetaProperty('name'));
+
+  // handle external references
+  if (metaRefOrigin !== baseURI) {
+    const hasMatchingMapping =
+      normalizedMapping.find((element: Element) =>
+        element.getMetaProperty('ref-fields')?.get('$ref')?.equals(metaRefFields?.$ref),
```
glowcloud: This might not be correct if we have different references that eventually point to the same thing. Not sure how else to compare them.

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,160 @@
...
+      ).length > 0;
+
+    if (!hasMatchingMapping) {
+      schemaElement.discriminator!.set('x-swagger-ui-normalized', false);
+    }
+    return;
+  }
+
+  // handle internal references that don't point to components/schemas/<SchemaName>
+  if (!metaSchemaName && metaRefFields) {
```
glowcloud: Not sure if this should be handled the same way as the external references.

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,160 @@
...
+  ({ storageField = 'x-normalized', baseURI = '' }: PluginOptions = {}) =>
+  (toolbox: Toolbox) => {
+    const { ancestorLineageToJSONPointer } = toolbox;
+    let storage: NormalizeStorage | undefined;
+
+    return {
+      visitor: {
+        OpenApi3_1Element: {
+          enter(element: OpenApi3_1Element) {
+            storage = new NormalizeStorage(element, storageField, 'discriminator-mapping');
```
glowcloud, line 47: I was testing normalization linking to Swagger Client and Swagger UI. Opening one operation would cause another not to be normalized. I'm not sure if this is due to the storage or if it is something else.

glowcloud: Fixed by setting to storage only if the element was actually normalized (`x-discriminator-mapping` field was added).

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+              return;
+            }
+
+            // skip if both oneOf and anyOf are present
+            if (isArrayElement(schemaElement.oneOf) && isArrayElement(schemaElement.anyOf)) {
+              return;
+            }
+
+            // skip if neither oneOf nor anyOf is present
+            if (!isArrayElement(schemaElement.oneOf) && !isArrayElement(schemaElement.anyOf)) {
```
glowcloud, line 84: This might need to be removed when we'll be handling `allOf` case.

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+            ancestors: [Element | Element[]],
+          ) {
+            // no Schema.discriminator field present
+            if (typeof schemaElement.discriminator === 'undefined') {
+              return;
+            }
+
+            const schemaJSONPointer = ancestorLineageToJSONPointer([
+              ...ancestors,
+              parent!,
```
robert-hebel-sb, line 69: won't passing an undefined parent cause a type error?

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+
+            // skip if neither oneOf nor anyOf is present
+            if (!isArrayElement(schemaElement.oneOf) && !isArrayElement(schemaElement.anyOf)) {
+              return;
+            }
+
+            const mapping =
+              schemaElement.discriminator.get('mapping') ?? new DiscriminatorMappingElement();
+            const normalizedMapping: DiscriminatorMappingElement = cloneDeep(mapping);
+            let normalizationResult = true;
```
robert-hebel-sb:
```suggestion
            let hasNormalizationResult = true;
```

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+            }
+
+            const schemaJSONPointer = ancestorLineageToJSONPointer([
+              ...ancestors,
+              parent!,
+              schemaElement,
+            ]);
+
+            // skip visiting this Schema Object if it's already normalized
+            if (storage!.includes(schemaJSONPointer)) {
```
robert-hebel-sb, line 74: Maybe we should check before whether storage is not being empty?

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+            let normalizationResult = true;
+
+            const items = isArrayElement(schemaElement.oneOf)
+              ? schemaElement.oneOf
+              : schemaElement.anyOf;
+
+            items!.forEach((item) => {
+              if (!isSchemaElement(item)) {
+                return;
+              }
```
robert-hebel-sb, line 99:
```suggestion
            const items = isArrayElement(schemaElement.oneOf)
              ? schemaElement.oneOf
              : (schemaElement.anyOf as ArrayElement);
              
            items.forEach((item) => {
              if (!isSchemaElement(item)) {
                return;
              }
```

packages/apidom-reference/src/dereference/strategies/openapi-3-1/visitor.ts
```diff
@@ -699,6 +701,95 @@ class OpenAPI3_1DereferenceVisitor {
...
+    const referencedElement = await visitAsync(memberElementSchema, visitor, {
+      keyMap,
+      nodeTypeGetter: getNodeType,
+    });
+
+    // remove referencing reference from ancestors lineage
+    directAncestors.delete(memberElementSchema);
+
+    // annotate MemberElement with referenced schema
+    const memberElementCopy: MemberElement = cloneDeep(memberElement);
```
char0n: shallow

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+        SchemaElement: {
+          leave(
+            schemaElement: SchemaElement,
+            key: string | number,
+            parent: Element | undefined,
+            path: (string | number)[],
+            ancestors: [Element | Element[]],
+          ) {
+            // no Schema.discriminator field present
+            if (typeof schemaElement.discriminator === 'undefined') {
```
char0n: Better to hook on expected shape, creating `isDiscriminatorElement` predicate and using it here.

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -0,0 +1,181 @@
...
+              return;
+            }
+
+            // skip if neither oneOf nor anyOf is present
+            if (!isArrayElement(schemaElement.oneOf) && !isArrayElement(schemaElement.anyOf)) {
+              return;
+            }
+
+            const mapping =
+              schemaElement.discriminator.get('mapping') ?? new DiscriminatorMappingElement();
```
char0n: new ObjectElement() might be more appropriate here

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
glowcloud: I think that the error might be expected here - the normalization plugins expect the `OpenAPI3_1Element` to be the root, otherwise they should fail. Similar case applies to https://github.com/swagger-api/apidom/pull/4952#discussion_r2158715388, as `storage` should be initialized when entering the `OpenAPI3_1Element`.

packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
glowcloud: Addressed in https://github.com/swagger-api/apidom/pull/4959

robert-hebel-sb (approved):

PR #4959, opened 2025-06-18
feat: add PoC of handling discriminator for allOf case in OpenAPI 3.1

glowcloud:


packages/apidom-ns-openapi-3-1/src/refractor/plugins/normalize-discriminator-mapping.ts
```diff
@@ -124,7 +146,7 @@ const plugin =
                     ?.get('$refBaseURI');
 
                   if (mappingValueSchemaRefBaseURI?.equals(metaRefFields?.$refBaseURI)) {
-                    normalizedMapping.set(toValue(mappingKey), cloneDeep(item));
+                    normalizedMapping.set(toValue(mappingKey), cloneShallow(item));
```
char0n, line 154: toValue not necessary

packages/apidom-ast/src/traversal/visitor.ts
```diff
@@ -531,6 +532,11 @@ export const visit = (
 
       // cycle detected; skipping over a sub-tree to avoid recursion
       if (detectCycles && ancestors.includes(node)) {
+        if (typeof detectCyclesCallback === 'function') {
+          // @ts-ignore
+          detectCyclesCallback(node, key, parent, path, ancestors);
```
robert-hebel-sb, line 537: adding types for visit options argument would fix that error

robert-hebel-sb (approved):

glowcloud:
> It seems there are many issues within typings. If resolving all TS errors exceeds the scope of the task, we can handle it in a separate ticket

Let's handle it separately - we have many similar cases with the typings, we could try addressing them as well.

PR #4964, opened 2025-06-27
perf: flatten sourceMap to improve memory usage

glowcloud:


packages/apidom-parser-adapter-yaml-1-2/src/syntactic-analysis/indirect/visitors/YamlAstVisitor.ts
```diff
@@ -212,18 +211,26 @@ class YamlAstVisitor {
     return null;
   }
 
+  /* eslint-disable no-param-reassign */
   private maybeAddSourceMap(node: unknown, element: Element): void {
```
char0n: node: unknown - type with `extends Node` from apidom-ast package

packages/apidom-core/src/predicates/index.ts
```diff
@@ -220,7 +206,14 @@ export const isPrimitiveElement: ElementPredicate<PrimitiveElement> = (
  * @public
  */
 export const hasElementSourceMap = <T extends Element>(element: T): boolean => {
-  return isSourceMapElement(element.meta.get('sourceMap'));
+  return (
```
char0n, line 213: Probably using a `Number.isInteger` is warranted here.

packages/apidom-core/src/traversal/findAtOffset.ts
```diff
@@ -29,9 +28,8 @@ class Visitor<T> {
       return undefined; // dive in
     }
 
-    const sourceMapElement = element.getMetaProperty('sourceMap');
-    const charStart = toValue(sourceMapElement.positionStart.get(2));
-    const charEnd = toValue(sourceMapElement.positionEnd.get(2));
+    const charStart = element.startIndex;
```
char0n: Align naming with new sourcemap fields.

packages/apidom-core/src/predicates/index.ts
char0n: And let's make this function safe - meaning it never throws. We can use `isElement` and only continue after truthy assertion.

packages/apidom-ns-api-design-systems/src/refractor/visitors/Visitor.ts
```diff
@@ -19,10 +19,14 @@ class Visitor {
   public copyMetaAndAttributes(from: Element, to: Element) {
     if (from.meta.length > 0 || to.meta.length > 0) {
       to.meta = deepmerge(to.meta, from.meta) as ObjectElement;
-      if (hasElementSourceMap(from)) {
-        // avoid deep merging of source maps
-        to.meta.set('sourceMap', from.meta.get('sourceMap'));
-      }
+    }
+    if (hasElementSourceMap(from)) {
+      to.startPositionRow = from.startPositionRow;
```
char0n:
```
assignSourceMap(to, from)
->
to.startPositionRow = from.startPositionRow
...
```

Possibly making it safe later. No parameter destricturing as it might be used in syntactic analysis or refracting, so we want to keep it fast.

packages/apidom-parser-adapter-json/src/syntactic-analysis/indirect/visitors/CstVisitor.ts
```diff
@@ -32,28 +30,16 @@ export const keyMap = {
...
-  }
-
   public readonly document = {
     enter: (node: TreeCursorSyntaxNode): JsonDocument => {
-      const position = CstVisitor.toPosition(node);
-
       return new JsonDocument({
         children: node.children,
-        position,
+        startPositionRow: node.startPositionRow,
```
char0n:
```js
return new JsonDocument(
  assignSourceMap({
    children: node.children,
    isMissing: node.isMissing,
  }, node)
);
```

char0n (approved):
Couple of comments to look at, but the overall direction is LEGIT

PR #4981, opened 2025-07-30
docs(ls): provide documentation with Language Service capabilities

glowcloud:
This PR adds documentation to the `apidom-ls` package, providing examples of the Language Service capababilities.


packages/apidom-ls/README.md
```diff
@@ -13,26 +13,749 @@ via [npm CLI](https://docs.npmjs.com/cli) by running the following command:
...
+  name: 'SB-API-052-title',
+  description: 'title MUST follow camelCase',
+  recommended: true,
+  code: 20001,
+  source: 'apilint',
+  message: 'title MUST follow camelCase',
+  given: '$.info.title',
+  givenFormat: 'JSONPATH',
+  severity: 1,
+  linterFunction: 'apilintValueCasing',
```
robert-hebel-sb, line 139: is it possible to create a custom linter function?

glowcloud: Yes, it would be something like this:
```js
const customLintFn = (element, expectedValues) => {
  if (element) {
    if (!isStringElement(element)) {
      return false;
    }

    return expectedValues.includes(toValue(element));
  }

  return true;
};

const titleEqualsRules = {
  code: 10063,
  source: "apilint",
  message: "title must be equal to one of the specified values",
  severity: 1,
  linterFunction: "apilintEqualsCustom",
  linterParams: ["Test 1", "Test 2"],
  given: "$.info.title",
  givenFormat: "JSONPATH",
  marker: "value",
  data: {},
};

customMetadata.linterFunctions = {
  asyncapi: {
    apilintEqualsCustom: customLintFn,
  },
};

customMetadata.rules = {
  asyncapi: {
    lint: [titleEqualsRules],
  },
};

const customContext = {
  metadata: customMetadata,
};

const languageService = getLanguageService(customContext);

const diagnostics = await languageService.doValidation(document);
```

Do we want to document this as well?

robert-hebel-sb: I see it being useful for extending apidom-ls capability, so I would go for it

lukaszzazulak (approved):

PR #4986, opened 2025-08-06
feat: add parameter validation rules

robert-hebel-sb:
Adds additional parameter validation rules

### DIFF
--- a/packages/apidom-ls/src/services/validation/linter-functions.ts
+++ b/packages/apidom-ls/src/services/validation/linter-functions.ts
@@ -801,13 +801,13 @@ export const standardLinterfunctions: FunctionItem[] = [
       const value = toValue(element);
 
       const filterSiblingsOAS2 = (
-        el: Element & { key: { content?: string }; content: { value?: string } },
+        el: Element & { key?: { content?: string }; content: { value?: string } },
       ) => isString(el) && el.key?.content === key && toValue(el.content.value) === value;
 
       const filterSiblingsOAS3 = (el: Element) =>
         isObject(el) && el.hasKey(key) && toValue(el.get(key)) === value;
 
-      const elements = filter((el) => {
+      const elements = filter((el: Element) => {
         const classes: string[] = toValue(el.getMetaProperty('classes', []));
 
         return (


packages/apidom-ls/src/config/openapi/parameter/lint/unique-name.ts
```diff
@@ -0,0 +1,21 @@
...
+  code: ApilintCodes.OPENAPI2_PARAMETER_FIELD_NAME_UNIQUE,
+  source: 'apilint',
+  message: 'Name must be unique among all parameters',
+  severity: DiagnosticSeverity.Error,
+  linterFunction: 'apilintPropertyUniqueSiblingValue',
+  linterParams: ['parameters', 'name'],
+  marker: 'key',
+  markerTarget: 'name',
+  target: 'name',
+  data: {},
```
lukaszzazulak, line 17: is the empty data needed ?

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -795,6 +795,30 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+      const value = toValue(element);
+
+      const filterSiblingsOAS2 = (
+        el: Element & { key: { content?: string }; content: { value?: string } },
+      ) => isString(el) && el.key?.content === key && toValue(el.content.value) === value;
+
+      const filterSiblingsOAS3 = (el: Element) =>
+        isObject(el) && el.hasKey(key) && toValue(el.get(key)) === value;
+
+      const elements = filter((el) => {
```
lukaszzazulak: the 'el' is any here right? can we do sth about it ?

robert-hebel-sb: right, I need to check why it's not picked up by eslint. Thanks for catching that. It's fixed

packages/apidom-ls/src/config/openapi/parameter/lint/unique-name.ts
robert-hebel-sb: I see it's a default in a case when we don't specify any quickfix rules

cka121 (approved):

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -795,6 +795,30 @@ export const standardLinterfunctions: FunctionItem[] = [
       return true;
     },
   },
+  {
+    functionName: 'apilintPropertyUniqueSiblingValue',
```
lukaszzazulak, line 799: sorry for the late reply, instead of having new linter, could the 'apilintPropertyUniqueValue'  be used ?

robert-hebel-sb: it wouldn't work for specifications with multiple parameters defined

PR #4987, opened 2025-08-08
feat(ls): migrate disallowing equivalent paths in OAS2

lukaszzazulak:
Migrate linting rule for disallowing equivalent paths in OAS2


packages/apidom-ls/src/config/openapi/path-template/lint/value--equivalent-not-allowed.ts
```diff
@@ -0,0 +1,17 @@
...
+import { OpenAPI2 } from '../../target-specs.ts';
+
+const valueEquivalentNotAllowedLint: LinterMeta = {
+  code: ApilintCodes.OPENAPI2_PATH_TEMPLATE_EQUIVALENT_NOT_ALLOWED,
+  source: 'apilint',
+  message: 'Equivalent paths are not allowed.',
+  severity: DiagnosticSeverity.Error,
+  linterFunction: 'apilintOpenAPIPathTemplateNoEquivalent',
+  marker: 'value',
+  targetSpecs: [...OpenAPI2],
```
robert-hebel-sb:
```suggestion
  targetSpecs: OpenAPI2,
```

packages/apidom-ls/src/config/openapi/path-template/lint/value--equivalent-not-allowed.ts
```diff
@@ -0,0 +1,17 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes.ts';
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+import { OpenAPI2 } from '../../target-specs.ts';
+
+const valueEquivalentNotAllowedLint: LinterMeta = {
+  code: ApilintCodes.OPENAPI2_PATH_TEMPLATE_EQUIVALENT_NOT_ALLOWED,
+  source: 'apilint',
+  message: 'Equivalent paths are not allowed.',
```
robert-hebel-sb:
```suggestion
  message: 'Equivalent paths are not allowed',
```

robert-hebel-sb (approved):

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -31,6 +33,16 @@ import {
 } from '../../utils/utils.ts';
 import { FunctionItem } from '../../apidom-language-types.ts';
 
+const getPathParams = (path: string) => {
+  const { ast } = parse(path);
```
glowcloud: We already import `parse` under a different name
```suggestion
  const { ast } = parsePathTemplate(path);
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -31,6 +33,16 @@ import {
...
+  const parts: any[] = [];
```
glowcloud:
```suggestion
  const parts: [string, string][] = [];
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1110,16 +1123,29 @@ export const standardLinterfunctions: FunctionItem[] = [
               oneOfParametersIsReferenceObject = true;
             }
             if (isParameterElement(parameter)) {
+              // separate array was needed, because parameterElements is extended in the loop below, but only global ones were needed
+              const name = isObject(parameter) && parameter.get('name');
```
glowcloud: We need to take into account only parameters that have `in: path`.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1144,7 +1170,19 @@ export const standardLinterfunctions: FunctionItem[] = [
         const resolvedPathTemplate = resolvePathTemplate(pathTemplate, pathTemplateResolveParams);
         const includesTemplateExpression = testPathTemplate(resolvedPathTemplate, { strict: true });
 
-        return !includesTemplateExpression || oneOfParametersIsReferenceObject;
+        const pathParams: string[] = getPathParams(element.toValue());
+
+        const everyPathParamInOperation = pathParams.every((pathParam) => {
```
glowcloud: For this comment no need to rewrite anything, I'm just wondering if it would be possible to do something like 
1. get path item level (global) parameters
2. for each operation
    - get its parameters
    - concat operation parameters with global parameters
    - test path template and save the result
3. check the results of testing path templates for each operation or `oneOfParametersIsReferenceObject` 

because it seems to me like we check if the template was / will be resolved twice. 

But I might be missing something and we most likely don't want to touch what was there before too much, so what we have seems fine.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1144,7 +1170,19 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          if (!globalParameterElementNames.includes(pathParam)) {
```
glowcloud: We could also do this:

```suggestion
          return (
            globalParameterElementNames.includes(pathParam) ||
            Object.values(operationLevelKeys).every((el) => el.includes(pathParam))
          );
```

packages/apidom-ls/src/services/validation/linter-functions.ts
lukaszzazulak: I've decided to refactor/simplify as you suggested, because there will be more changes needed soon, and it will be easier to implement them, please have a look now.

packages/apidom-ls/src/services/validation/linter-functions.ts
lukaszzazulak: I've refactored, please have a look now.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,74 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        const pathLevelParameterElements: Element[] = [];
         const pathItemParameterElements = pathItemElement.get('parameters');
         if (isArrayElement(pathItemParameterElements)) {
           pathItemParameterElements.forEach((parameter) => {
             if (isReferenceElement(parameter) && !oneOfParametersIsReferenceObject) {
               oneOfParametersIsReferenceObject = true;
             }
             if (isParameterElement(parameter)) {
-              parameterElements.push(parameter);
+              if (isObject(parameter)) {
```
glowcloud:
```suggestion
              if (isParameterElement(parameter) && isObject(parameter)) {
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,74 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+                const inParam = parameter.get('in');
+                const nameParam = parameter.get('name');
+                if (inParam && inParam.content === 'path' && nameParam && nameParam.content) {
```
glowcloud: We also check the same thing later after concatenating the arrays (lines 1145-1147) - perhaps we can remove it here and just check it there?

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,74 @@ export const standardLinterfunctions: FunctionItem[] = [
...
           }
         });
 
-        const pathTemplate = toValue(element);
-        const resolvedPathTemplate = resolvePathTemplate(pathTemplate, pathTemplateResolveParams);
-        const includesTemplateExpression = testPathTemplate(resolvedPathTemplate, { strict: true });
-
-        return !includesTemplateExpression || oneOfParametersIsReferenceObject;
+        return (
+          (includesTemplateExpression.length
```
glowcloud:
```suggestion
          (includesTemplateExpression.length > 0 && includesTemplateExpression.every((bool) => !bool)) || oneOfParametersIsReferenceObject)
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,74 @@ export const standardLinterfunctions: FunctionItem[] = [
...
             }
-          }
-        });
-
-        const pathTemplateResolveParams: { [key: string]: 'placeholder' } = {};
+            const pathTemplateResolveParams: { [key: string]: 'placeholder' } = {};
+            pathLevelParameterElements
+              .concat(currentOperationLevelParameterElements)
+              .forEach((parameter) => {
+                if (isObject(parameter)) {
```
glowcloud: We already have this condition for path-level parameters (line 1114) - perhaps we should remove it from there or remove it from here and add it to operation-level parameters (line 1135)?

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,66 @@ export const standardLinterfunctions: FunctionItem[] = [
...
           }
         });
 
-        const pathTemplate = toValue(element);
-        const resolvedPathTemplate = resolvePathTemplate(pathTemplate, pathTemplateResolveParams);
-        const includesTemplateExpression = testPathTemplate(resolvedPathTemplate, { strict: true });
-
-        return !includesTemplateExpression || oneOfParametersIsReferenceObject;
+        return (
+          (includesTemplateExpression.length > 0
```
glowcloud: We don't need a ternary operator here
```ts
includesTemplateExpression.length > 0 && includesTemplateExpression.every((bool) => !bool)
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1099,52 +1100,66 @@ export const standardLinterfunctions: FunctionItem[] = [
...
         const pathItemParameterElements = pathItemElement.get('parameters');
         if (isArrayElement(pathItemParameterElements)) {
           pathItemParameterElements.forEach((parameter) => {
             if (isReferenceElement(parameter) && !oneOfParametersIsReferenceObject) {
               oneOfParametersIsReferenceObject = true;
             }
-            if (isParameterElement(parameter)) {
-              parameterElements.push(parameter);
-            }
+            pathLevelParameterElements.push(parameter);
```
glowcloud: Do we not want to check if this is a parameter element here?
```ts
if (isParameterElement(parameter))
```

glowcloud (approved):

PR #4991, opened 2025-08-14
Feat/swg 15853 port validation security

lukaszzazulak:
Migrate linting rule for disallowing unused schemes and unresolved scopes.


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1133,4 +1133,98 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+      const isFirstOccurrence = (currentKey: string, allKeys: unknown[]) => {
+        const normalize = (x: string) => x.replace(PATH_TEMPLATES_REGEX, '~~');
+        const currentKeyNormalized = normalize(currentKey);
+        const firstIndex = allKeys.findIndex(
+          (e) => typeof e === 'string' && normalize(e) === currentKeyNormalized,
+        );
+
+        return allKeys[firstIndex] === currentKey;
+      };
+      const paths = element.parent.parent;
```
robert-hebel-sb:
```suggestion
      const paths = element.parent?.parent;
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1133,4 +1133,98 @@ export const standardLinterfunctions: FunctionItem[] = [
       return true;
     },
   },
+  {
+    functionName: 'apilintOpenAPIPathTemplateNoEquivalent',
+    function: (element: Element): boolean => {
+      const PATH_TEMPLATES_REGEX = /\{[^}]+\}/g;
+      const isFirstOccurrence = (currentKey: string, allKeys: unknown[]) => {
```
glowcloud, line 1140: We use `openapi-path-templating` package for testing if the path is valid. It also provides functions for checking if the [paths are equivalent](https://github.com/swaggerexpert/openapi-path-templating?tab=readme-ov-file#equivalence) - perhaps we could use that instead?

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1133,4 +1133,98 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+      return isStringElement(element) && isObject(paths)
+        ? isFirstOccurrence(element.toValue(), paths.keys())
+        : true;
+    },
+  },
+  {
+    functionName: 'apilintSecurityScopeResolved',
+    function: (element: Element): boolean => {
+      const api = root(element);
+      const securityDefinitions = typeof api.get === 'function' && api.get('securityDefinitions');
```
glowcloud, line 1160: Would using `isObject` not work here?
```suggestion
      const securityDefinitions = isObject(api) && api.get('securityDefinitions');
```

packages/apidom-ls/src/config/codes.ts
```diff
@@ -667,6 +668,7 @@ enum ApilintCodes {
   OPENAPI2_PATH_TEMPLATE = 3040000,
   OPENAPI2_PATH_TEMPLATE_VALUE_WELL_FORMED = 3040100,
   OPENAPI2_PATH_TEMPLATE_VALUE_VALID,
+  OPENAPI2_PATH_TEMPLATE_EQUIVALENT_NOT_ALLOWED,
```
glowcloud, line 671: Nitpick: should we start each path template code name for its value the same way with `OPENAPI2_PATH_TEMPLATE_VALUE_`, so here `OPENAPI2_PATH_TEMPLATE_VALUE_EQUIVALENT_NOT_ALLOWED`?

lukaszzazulak:
Closing this one because it contains unwanted changes form different branch.

PR #4992, opened 2025-08-14
feat(ls): add schema validation rules

robert-hebel-sb:


cka121 (approved):

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -544,6 +612,25 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+      }
+
+      const minimum = toValue(elementParent.get(minProperty));
+      const maximum = toValue(elementParent.get(maxProperty));
+
+      if (typeof minimum !== 'number' || typeof maximum !== 'number') {
+        return true;
+      }
+
+      return minimum <= maximum;
```
glowcloud, line 656: In case of `minimum` and `maximum` there also exist `exclusiveMinimum` and `exclusiveMaximum` fields - should they also be taken into account for this?

robert-hebel-sb: good catch! changed to take exclusiveMinimum/exclusiveMaximum into consideration

packages/apidom-ls/src/config/common/schema/lint/enum--type.ts
```diff
@@ -0,0 +1,20 @@
...
+  code: ApilintCodes.SCHEMA_ENUM_TYPE,
+  source: 'apilint',
+  message: "enum value should conform to its schema's type",
+  severity: DiagnosticSeverity.Warning,
+  linterFunction: 'apilintArrayOfType',
+  linterParams: ['', false, true],
+  marker: 'value',
+  target: 'enum',
+  data: {},
+  targetSpecs: [...OpenAPI2, ...OpenAPI3],
```
glowcloud: Could you verify if this will also apply to OpenAPI 3.1.0, considering that the [target spec](https://github.com/swagger-api/apidom/blob/e8570c44f754347dc450c10f95fb239e92cd8f2e/packages/apidom-ls/src/config/openapi/target-specs.ts#L17) also includes it? If so, could you also verify if this rule will work for multiple types, which can be defined in OpenAPI 3.1.0, e.g.
```yaml
type:
  - number
  - boolean
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -559,6 +627,35 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        minProperty === 'minimum' &&
+        maxProperty === 'maximum' &&
+        (toValue(elementParent.get('exclusiveMinimum')) ||
+          toValue(elementParent.get('exclusiveMaximum')));
+
+      if (typeof minimum !== 'number' || typeof maximum !== 'number') {
+        return true;
+      }
+
+      if (isExclusiveMinimumMaximumEnabled) {
```
glowcloud, line 652: Would we need to verify if this is a `boolean`, if someone can provide something else? Or perhaps it will be caught by a different rule?

packages/apidom-ls/src/services/validation/linter-functions.ts
glowcloud: One more thing here, similar to `enum` rule - if this rule also applies to OpenAPI 3.1.0, then we would probably need to handle it differently, as `exclusiveMinimum` and `exclusiveMaximum` are numbers there.

robert-hebel-sb: in this PR I aim to bring compatibility between validation rules used in editorV4 and editorV5. We should create a separate PR for 3.1 additional validation

packages/apidom-ls/src/services/validation/linter-functions.ts
robert-hebel-sb: current rule applies only to 3.0 which specify this property to be bool, rule for 3.1 will be addressed seperatly

packages/apidom-ls/src/config/common/schema/lint/enum--type.ts
robert-hebel-sb: rules for 3.1 will be handled in separate PR

PR #4995, opened 2025-08-19
Feat/swg 15853 port validation security clean

lukaszzazulak:
Migrate linting rule for disallowing unused schemes and unresolved scopes.


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1157,4 +1157,78 @@ export const standardLinterfunctions: FunctionItem[] = [
...
     },
   },
+  {
+    functionName: 'apilintSecurityScopeResolved',
+    function: (element: Element): boolean => {
+      const api = root(element);
+      const securityDefinitions = isObject(api) && api.get('securityDefinitions');
+      if (!securityDefinitions || !isObject(securityDefinitions)) return true;
+
+      const securityRequirements = element.toValue() as unknown;
```
glowcloud: Perhaps we could do this to make sure it's an Object Element?
```suggestion
      const scopeExists: boolean[] = [];
      if (isObject(element)) {
         const securityRequirements = element.toValue()
      }
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1157,4 +1157,78 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+      const scopeExists: boolean[] = [];
+      if (securityRequirements) {
+        for (const [schemeName, scopesFromSecurity] of Object.entries(securityRequirements)) {
+          if (Array.isArray(scopesFromSecurity)) {
+            scopeExists.push(
+              scopesFromSecurity.every((scopeFromSecurity: string) => {
```
glowcloud: Would it make sense to extract this to a separate function so that we don't need to check the same thing in every loop, e.g.
```ts
      const hasScope = (scopesFromSecurity: string[], schemeName: string) => {
        const oneSecurityDefinition = securityDefinitions.get(schemeName);
        if (!oneSecurityDefinition) return true;

        const oneSecurityDefinitionScopes = oneSecurityDefinition.get('scopes');
        if (!oneSecurityDefinitionScopes) return true;

        return scopesFromSecurity.every(
          (scope: string) => !!oneSecurityDefinitionScopes.get(scope),
        );
      };

     // ...
     scopeExists.push(hasScope(scopesFromSecurity, schemeName))
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1157,4 +1157,78 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+                    });
+                  }
+                }
+              });
+            }
+          });
+        }
+        // Check if schemeName is used in any security object
+        for (const secObj of securityObjects) {
+          if (isObject(secObj)) {
```
glowcloud: I think we won't need to check this if we define the type for `securityObjects` as `ObjectElement[]`, as we also check this type when pushing to this array - so we should not have any non-Object Elements.

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1157,4 +1157,78 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+    functionName: 'apilintSecuritySchemeUsed',
+    function: (element: Element): boolean => {
+      if (element && element.parent && isMember(element.parent)) {
+        const schemeName: string = element.parent.toValue().key;
+        const api = root(element);
+        const securityObjects: Element[] = [];
+        const globalSecurity = isObject(api) && api.get('security');
+        if (globalSecurity && isArray(globalSecurity)) {
+          globalSecurity.forEach((secObj) => {
+            if (isObject(secObj)) securityObjects.push(secObj);
```
glowcloud: Would it be better if instead of pushing into the array and then iterating through it again to find the `schemeName`, we would just have some variable, which would check it during the iteration here and in the other loop? E.g.
```ts
let hasSchemeName = false;
// ...
if (isObject(secObj) && secObj.hasKey(schemeName)) hasSchemeName = true
// ...
return hasSchemeName
```

packages/apidom-ls/src/config/openapi/security-scheme/lint/keys--used.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const keysUsedLint: LinterMeta = {
+  code: ApilintCodes.SECURITY_SCHEME_USED,
+  source: 'apilint',
+  message:
+    'Security Scheme was defined but never used. To apply security, use the `security` section in operations or on the root level of your API definition',
+  severity: DiagnosticSeverity.Warning,
+  linterFunction: 'apilintSecuritySchemeUsed',
+  marker: 'key',
+  targetSpecs: [...OpenAPI2, ...OpenAPI3],
```
robert-hebel-sb:
```suggestion
import { OpenAPI2, OpenAPI30 } from '../../target-specs.ts';

const keysUsedLint: LinterMeta = {
  code: ApilintCodes.SECURITY_SCHEME_USED,
  source: 'apilint',
  message:
    'Security Scheme was defined but never used. To apply security, use the `security` section in operations or on the root level of your API definition',
  severity: DiagnosticSeverity.Warning,
  linterFunction: 'apilintSecuritySchemeUsed',
  marker: 'key',
  targetSpecs: [...OpenAPI2, ...OpenAPI30],
```



In the current scope lets focus only on 2.0 and 3.0 definitions. 3.1 Parity will be handled seperatly

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1219,4 +1220,75 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+          for (const secObj of globalSecurity) {
+            if (secObj.hasKey(schemeName)) {
+              return true;
+            }
+          }
+        }
+        const paths: ObjectElement = isObject(api) && api.get('paths');
+        return !!paths.findElements(
+          (e) => {
+            if (isObjectElement(e)) {
```
glowcloud:
```suggestion
            if (isObject(e) && e.hasKey('security')) {
```

packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1219,4 +1220,75 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+              if (e.hasKey('security')) {
+                const opSecurity = e.get('security');
+                if (isArrayElement(opSecurity)) {
+                  return !!opSecurity.findElements(
+                    (securityRequirementObject) =>
+                      isObjectElement(securityRequirementObject) &&
```
glowcloud: I think `isObject` and `isObjectElement` are equivalent, so we don't need to import both of them. 

Similarly in other places we have e.g. `isArray` and `isArrayElement` both, in existing functions as well. We could probably refactor them in the future to use only one of them, if they are truly equivalent.

```suggestion
                      isObject(securityRequirementObject) &&
```

glowcloud (approved):
Looks good, just small suggestions.

PR #4998, opened 2025-08-26
feat(ls): add validation rules for formData

robert-hebel-sb:


packages/apidom-ls/src/config/openapi/operation/lint/parameters-type-consumes--required.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import ApilintCodes from '../../../codes.ts';
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+import { OpenAPI2 } from '../../target-specs.ts';
+
+const parametersTypeConsumesRequiredLint: LinterMeta = {
+  code: ApilintCodes.OPENAPI2_OPERATION_FIELD_PARAMETERS_TYPE_CONSUMES_REQUIRED,
+  source: 'apilint',
+  message:
+    'Operations with parameters of "type: file" must include "multipart/form-data" in their "consumes" property',
```
glowcloud: This one seems to contradict the specification:
parameter_consumes

packages/apidom-ls/src/config/openapi/operation/lint/parameters-type-consumes--required.ts
```diff
@@ -0,0 +1,27 @@
+import { DiagnosticSeverity } from 'vscode-languageserver-types';
+
+import ApilintCodes from '../../../codes.ts';
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+import { OpenAPI2 } from '../../target-specs.ts';
+
+const parametersTypeConsumesRequiredLint: LinterMeta = {
```
glowcloud, line 7: Can we also verify that this rule will be correct for global `consumes`? And that it will also work if global `consumes` is overriden on operation-level?

robert-hebel-sb: good point, added fallback check + tests

packages/apidom-ls/src/config/openapi/operation/lint/parameters-type-consumes--required.ts
robert-hebel-sb: good catch! fixed

PR #5001, opened 2025-09-02
fix: fallback to a default version when a spec version is not found

robert-hebel-sb:


packages/apidom-ls/src/utils/utils.ts
```diff
@@ -166,9 +166,9 @@ export function setMetadataMap(
   }
 }
 
-export function getSpecVersion(root: Element): string {
+export function getSpecVersion(root: Element, defaultVersion?: string): string {
   const el = find((e) => toValue(e.getMetaProperty('classes', []).includes('spec-version')), root);
-  return el ? toValue(el) : '';
+  return (el ? toValue(el) : defaultVersion) ?? '';
```
glowcloud: We could separate this:
```suggestion
  if (el) {
    return toValue(el)
  }
  return defaultVersion ?? '';
```

glowcloud (approved):

PR #5003, opened 2025-09-04
feat(ls): add option to pass refractor plugins

glowcloud:


robert-hebel-sb:
great work! 🚀 
Is it something we would like to include in the documentation?

PR #5015, opened 2025-09-19
fix(ls): external refs should not be validated

robert-hebel-sb:
fixes unwanted validation errors against external refs


packages/apidom-ls/src/config/common/schema/lint/$ref-3-0--request-bodies.ts
```diff
@@ -16,7 +16,7 @@ const $ref3RequestBodiesLint: LinterMeta = {
     {
       targets: [{ path: '$ref' }],
       function: 'apilintValueRegex',
-      params: ['^(?!.*#/components/schemas).*$'],
+      params: ['^(?!.*#/components/schemas)(?!https?://).*$'],
```
glowcloud: Would this not mean that only external references starting with `http(s)://` are not validated? We could have any URI here, some examples here:

refs

robert-hebel-sb: good point! Changed to validating local references only

glowcloud (approved):

PR #5039, opened 2025-10-30
test(ns-asyncapi-3): add tests for AsyncAPI 3 namespace

glowcloud:
Add tests and fixes related to them.

TODO:
- [x] overall refractor tests
- [x] elements tests
- [x] perf tests
- [x] replace empty element plugin tests
  - [x] tests
  - [x] snapshots 
- [x] predicates tests
  - [x] copy tests from AsyncAPI 2
  - [x] isMultiFormatSchemaElement tests
  - [x] missing tests for predicates 
- [x] verify some elements/visitors (e.g. `MultiFormatSchema`) => verified and adjusted based on snapshots, rest should be done in namespace PR

### DIFF
--- a/packages/apidom-ns-asyncapi-3/src/refractor/plugins/replace-empty-element.ts
+++ b/packages/apidom-ns-asyncapi-3/src/refractor/plugins/replace-empty-element.ts
@@ -197,549 +197,549 @@ import { getNodeType } from '../../traversal/visitor.ts';
 const isEmptyElement = (element: unknown): element is StringElement =>
   isStringElement(element) && includesClasses(['yaml-e-node', 'yaml-e-scalar'], element);
 
-const schema: Record<string, any> = {
+const schema: Record<string, unknown> = {
   AsyncApi3Element: {
-    asyncapi(...args: any[]) {
+    asyncapi(...args: string[]) {
       return new AsyncApiVersionElement(...args);
     },
-    identifier(...args: any[]) {
+    identifier(...args: string[]) {
       return new IdentifierElement(...args);
     },
-    info(...args: any[]) {
+    info(...args: Record<string, unknown>[]) {
       return new InfoElement(...args);
     },
-    servers(...args: any[]) {
+    servers(...args: Record<string, unknown>[]) {
       return new ServersElement(...args);
     },
-    defaultContentType(...args: any[]) {
+    defaultContentType(...args: string[]) {
       return new DefaultContentTypeElement(...args);
     },
-    channels(...args: any[]) {
+    channels(...args: Record<string, unknown>[]) {
       return new ChannelsElement(...args);
     },
-    components(...args: any[]) {
+    components(...args: Record<string, unknown>[]) {
       return new ComponentsElement(...args);
     },
-    operations(...args: any[]) {
+    operations(...args: Record<string, unknown>[]) {
       return new OperationsElement(...args);
     },
   },
 
   InfoElement: {
-    contact(...args: any[]) {
+    contact(...args: Record<string, unknown>[]) {
       return new ContactElement(...args);
     },
-    license(...args: any[]) {
+    license(...args: Record<string, unknown>[]) {
       return new LicenseElement(...args);
     },
-    tags(...args: any[]) {
+    tags(...args: ConstructorParameters<typeof TagsElement>) {
       return new TagsElement(...args);
     },
-    externalDocs(...args: any[]) {
+    externalDocs(...args: Record<string, unknown>[]) {
       return new ExternalDocumentationElement(...args);
     },
   },
 
   ServersElement: {
-    '[key: *]': function key(...args: any[]) {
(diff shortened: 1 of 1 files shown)


packages/apidom-ns-asyncapi-3/src/refractor/specification.ts
```diff
@@ -607,11 +730,11 @@ const specification = {
         MultiFormatSchema: {
           $visitor: MultiFormatSchemaVisitor,
           fixedFields: {
-            $shemaFormat: { $ref: '#/visitors/document/objects/SchemaFormat' },
-            $schema: { $ref: '#/visitors/document/objects/Schema' },
-            properties: { $ref: '#/visitors/document/objects/MultiFormatSchema' },
+            schemaFormat: { $ref: '#/visitors/value' },
+            schema: { $ref: '#/visitors/document/objects/Schema' },
```
glowcloud: We might need to think more about the visitor for this, as schema can be anything based on what's defined in `schemaFormat`, as per the [documentation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject).

packages/apidom-ns-asyncapi-3/src/refractor/visitors/async-api-3/multiFormatSchema/index.ts
```diff
@@ -26,19 +32,30 @@ class MultiFormatSchemaVisitor extends Mixin(FixedFieldsVisitor, FallbackVisitor
...
     this.specPath = always(['document', 'objects', 'MultiFormatSchema']);
     this.canSupportSpecificationExtensions = true;
   }
 
   ObjectElement(objectElement: ObjectElement) {
-    this.element = new MultiFormatSchemaElement();
-
-    return FixedFieldsVisitor.prototype.ObjectElement.call(this, objectElement);
-  }
+    const result = FixedFieldsVisitor.prototype.ObjectElement.call(this, objectElement);
```
glowcloud, line 41: Handling of Multi Format Schema based on [MessageVisitor](https://github.com/swagger-api/apidom/blob/e08fdf4c8a8070453c33066cc50f4e949fc4bd20/packages/apidom-ns-asyncapi-2/src/refractor/visitors/async-api-2/message/index.ts#L38) from AsyncAPI 2 namespace.

packages/apidom-ns-asyncapi-3/src/refractor/predicates.ts
```diff
@@ -14,6 +20,22 @@ export const isAsyncApiExtension = (element: MemberElement): boolean => {
   return isStringElement(element.key) && startsWith('x-', toValue(element.key));
 };
 
+/**
+ * @public
+ */
+export interface MultiFormatSchemaLikeElement extends ObjectElement {
+  hasKey: (value: 'schemaFormat') => true;
```
glowcloud, line 38: Question: should we be checking for both `schemaFormat` AND `schema` presence, either `schemaFormat` OR `schema` presence or is just checking for `schemaFormat` enough? This is for cases where we can have `ReferenceElement`, `SchemaElement` or `MultiFormatSchemaElement`. Currently recognising each element:
- ReferenceElement - recognised when `$ref` is present, e.g.:
```yaml
$ref: <value>


$ref: <value>
prop: <value>

$ref: <value>
schemaFormat: <value>
```
- MultiFormatSchemaElement - recognised when `schemaFormat` is present, e.g.:
```yaml
schemaFormat: <value>

schemaFormat: <value>
schema: <value>

schemaFormat: <value>
prop: <value>
```
- SchemaElement - defaulted to if element was not recognised as `ReferenceElement` or `MultiFormatSchemaElement`, e.g.:
```yaml
prop: <value>

schema: <value>
```

packages/apidom-ns-asyncapi-3/src/refractor/specification.ts
```diff
@@ -205,38 +324,38 @@ const specification = {
           $visitor: OperationVisitor,
           fixedFields: {
             action: { $ref: '#/visitors/value' },
-            channel: { $ref: '#/visitors/value' },
+            channel: OperationChannelVisitor,
```
glowcloud, line 339: `Operation.channel` is a `ReferenceElement` but I added a separate visitor for it, so that we can decorate it with metadata (`referenced-element: channel`). Not sure if we can do it in another way, if we want to have separate visitors for  `Operation.channel` and `OperationReply.channel`, and if we want the metadata for them to be different, as currently they both use this visitor.

packages/apidom-ns-asyncapi-3/test/perf/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/no-var-requires": 0
```
robert-hebel-sb: do we still need it?

glowcloud: We can remove it in other packages in the future as well.

PR #5040, opened 2025-11-03
feat(ls): add validation for empty responses object

glowcloud:
Adds validation for empty Responses Object. If the Responses Object has at least one non-specification extension field, there will be no validation error.

- no fields => error shown
```
responses:
```

- specification extension field only => error shown
```
responses:
   x-extension: 123
```

- incorrect HTTP Status Code => no error
```
responses:
   test:
      description: ok
```

- correct HTTP Status Code => no error
```
responses:
   '200':
      description: ok
```


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -1479,4 +1479,17 @@ export const standardLinterfunctions: FunctionItem[] = [
...
   },
+  {
+    functionName: 'apilintOpenAPIEmptyResponses',
+    function: (element: Element): boolean => {
+      if (element && isObject(element)) {
+        if (!element.keys() || element.keys().length === 0) {
+          return false;
+        }
+
+        return element.keys().some((k) => typeof k !== 'string' || !k.startsWith('x-'));
```
robert-hebel-sb, line 1490: won't it trigger it if at least one response won't pass the check?
btw. should we take into consideration `default` response?

glowcloud: We only trigger the validation rule if the function returns false. If we find any key here that does not start with `x-`, the function will return true, so it shouldn't be triggered. `default` is also taken into account as any key that does not start with `x-`.

PR #5047, opened 2025-11-13
feat(parser-adapter-asyncapi-yaml-3): add package for parser adapter …

lukaszzazulak:
Parser adapter for asyncapi yaml 3

PR #5048, opened 2025-11-17
feat: parser adapter for json in asyncapi3

cka121:
Parser adapter for asyncapi JSON 3


package.json
```diff
@@ -42,7 +42,8 @@
   },
   "homepage": "https://github.com/swagger-api/apidom#readme",
   "workspaces": [
-    "./packages/*"
+    "./packages/*",
+    "./packages/apidom-parser-adapter-asyncapi-json-3"
```
glowcloud: Will the current path `./packages/*` not include it?
```suggestion
```

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -450,6 +450,10 @@ export class DefaultValidationService implements ValidationService {
 
     let processedText;
     const nameSpace = await findNamespace(text, this.settings?.defaultContentLanguage);
+    // TODO: Turned off validation, because we will implement it in the future.
```
glowcloud, line 453: Same comment as in https://github.com/swagger-api/apidom/pull/5047#discussion_r2533050652, question if we need to have this now.

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,94 @@
+# @swagger-api/apidom-parser-adapter-asyncapi-json-3
+
+`@swagger-api/apidom-parser-adapter-asyncapi-json-3` is a parser adapter for following AsyncAPI specification versions defined in [JSON format](https://www.json.org/json-en.html):
+
+
```
glowcloud: Let's add which versions are supported:
```suggestion
- [AsyncAPI 3.0.0 specification](https://github.com/asyncapi/spec/blob/v3.0.0/spec/asyncapi.md)
```

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,94 @@
...
+```js
+import { parse, detect } from '@swagger-api/apidom-parser-adapter-asyncapi-json-3';
+
+// detecting
+await detect('{"asyncapi": "3.0.0"}'); // => true
+await detect('{"asyncapi": "3.0.1"}'); // => true
+await detect('test'); // => false
+
+// parsing
+const parseResult = await parse('{"asyncapi": "3.6.0"}', { sourceMap: true });
```
glowcloud: Let's use version 3.0.0 (or 3.0.1, as I see that's what we have in `Indirect usage` section)
```suggestion
const parseResult = await parse('{"asyncapi": "3.0.0"}', { sourceMap: true });
```

packages/apidom-parser-adapter-asyncapi-json-3/test/adapter.ts
```diff
@@ -0,0 +1,94 @@
...
+    specify('should detect patch version bump', async function () {
+      assert.isTrue(await adapter.detect('{"asyncapi": "3.6.1"}'));
+    });
+
+    specify('should detect minor and patch version bump', async function () {
+      assert.isTrue(await adapter.detect('{"asyncapi": "3.25.1"}'));
+    });
+  });
+
+  context('given definition in YAML 1.3 format', function () {
```
glowcloud, line 33:
```suggestion
  context('given definition in YAML 1.2 format', function () {
```

packages/apidom-parser-adapter-asyncapi-json-3/test/adapter.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    specify('should reject invalid version ranges', function () {
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "2.0.0"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "2.1.0"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "2.1.1"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "2.5.0"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "2.6.01"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "3.6.01"'));
+      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "3.6.01"'));
```
glowcloud: Duplicate of the test above, possibly should test something like this, looking at the tests from Async 2 parser:
```suggestion
      assert.isFalse(adapter.detectionRegExp.test('"asyncapi": "3.06.0"'));
```

packages/apidom-parser-adapter-asyncapi-json-3/test/__snapshots__/adapter.mjs.snap
```diff
@@ -0,0 +1,985 @@
...
+                        (StringElement))
+                      (MemberElement
+                        (StringElement)
+                        (StringElement))
+                      (MemberElement
+                        (StringElement)
+                        (StringElement))
+                      (MemberElement
+                        (StringElement)
+                        (ObjectElement
```
glowcloud, line 154: Not sure if I'm looking at this snapshot correctly but if so, we might have an issue with refractor for OAuthFlows (or just `flows` field in SecurityScheme). If this corresponds to this part of the fixture:
https://github.com/swagger-api/apidom/blob/841ae37270bc3bd51ecff9a558d9c81a8abf939e/packages/apidom-ns-asyncapi-3/test/refractor/fixtures/asyncapi-3-0-0.json#L59-L64
then we should have OAuthFlowsElement here and its MemberElements should have OAuthFlowElement instead of ObjectElement, comparing it to [this snapshot](https://github.com/swagger-api/apidom/blob/feat/OSS-248-parser-adapter-for-json-async3/packages/apidom-ns-asyncapi-3/test/refractor/elements/OAuthFlows/__snapshots__/index.ts.snap).

If this is verified by others, we should probably handle it separately as a bug and update the snapshots with that fix.

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,94 @@
+# @swagger-api/apidom-parser-adapter-asyncapi-json-3
+
+`@swagger-api/apidom-parser-adapter-asyncapi-json-3` is a parser adapter for following AsyncAPI specification versions defined in [JSON format](https://www.json.org/json-en.html):
+
+
+
+Under the hood this adapter uses [@swagger-api/apidom-parser-adapter-json](https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser-adapter-json)
+to parse a source string into generic ApiDOM in [base ApiDOM namespace](https://github.com/swagger-api/apidom/tree/main/packages/apidom#base-namespace)
```
glowcloud: Noticed that we have this wrong while working on fixing URLs in the docs for other packages. Some links that point to the code might also need adjustments - they should point to the correct line.
```suggestion
to parse a source string into generic ApiDOM in [base ApiDOM namespace](https://github.com/swagger-api/apidom/tree/main/packages/apidom-core#base-namespace)
```

package.json
cka121: Probably a leftover from the lint issue fix. I shall remove it

packages/apidom-ls/src/services/validation/validation-service.ts
cka121: If it's not required at this point we can remove it. Should I remove it ?

packages/apidom-parser-adapter-asyncapi-json-3/test/__snapshots__/adapter.mjs.snap
glowcloud: The bug was documented and will be handled separately.

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+### mediaTypes
+
+Defines list of media types that this parser adapter recognizes.
+
+```js
+[
+  'application/vnd.aai.asyncapi;version=3.0.0',
+  'application/vnd.aai.asyncapi+json;version=3.0.0',
+  'application/vnd.aai.asyncapi+yaml;version=3.0.0',
```
glowcloud: Let's remove the YAML versions here as this is only JSON parser.
```suggestion
```

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,95 @@
...
+  'application/vnd.aai.asyncapi;version=3.0.1',
+  'application/vnd.aai.asyncapi+json;version=3.0.1',
+  'application/vnd.aai.asyncapi+yaml;version=3.0.1',
```
glowcloud:
```suggestion
```

packages/apidom-parser-adapter-asyncapi-json-3/README.md
```diff
@@ -0,0 +1,95 @@
...
+]
+```
+
+### detect
+
+[Detection](https://github.com/swagger-api/apidom/blob/main/packages/apidom-parser-adapter-asyncapi-json-3/src/adapter.ts#L13) is based on a regular expression matching required AsyncApi 3.0.0 specification symbols in JSON format.
```
glowcloud: This should point to the `detect` function:

```suggestion
[Detection](https://github.com/swagger-api/apidom/blob/main/packages/apidom-parser-adapter-asyncapi-json-3/src/adapter.ts#L18) is based on a regular expression matching required AsyncApi 3.0.0 specification symbols in JSON format.
```

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -450,6 +450,10 @@ export class DefaultValidationService implements ValidationService {
...
+    if (nameSpace.namespace === 'asyncapi' && nameSpace.version === '3.0.0') {
```
glowcloud: Do we need to turn it off for the 3.0.1 version as well?

glowcloud (approved):
Looks good now, just small nitpicks.

packages/apidom-ls/src/services/validation/validation-service.ts
cka121: Okay

PR #5051, opened 2025-11-20
feat(apidom-reference): resolve references async v3

lukaszzazulak:
Resolve references for async v3


packages/apidom-ls/src/parser-factory.ts
```diff
@@ -6,10 +6,12 @@ import * as openapi3_1AdapterJson from '@swagger-api/apidom-parser-adapter-opena
 import * as openapi3_1AdapterYaml from '@swagger-api/apidom-parser-adapter-openapi-yaml-3-1';
 import * as asyncapi2AdapterJson from '@swagger-api/apidom-parser-adapter-asyncapi-json-2';
 import * as asyncapi2AdapterYaml from '@swagger-api/apidom-parser-adapter-asyncapi-yaml-2';
+import * as asyncapi3AdapterYaml from '@swagger-api/apidom-parser-adapter-asyncapi-yaml-3';
```
glowcloud, line 10: Should we add JSON parser adapter as well?

packages/apidom-ls/src/parser-factory.ts
```diff
@@ -39,7 +41,11 @@ export async function parse(
   const text: string = typeof textDocument === 'string' ? textDocument : textDocument.getText();
   let result;
   const contentLanguage = await findNamespace(text, defaultContentLanguage);
-  if (contentLanguage.namespace === 'asyncapi' && contentLanguage.format === 'JSON') {
+  if (
+    contentLanguage.namespace === 'asyncapi' &&
+    (contentLanguage.version?.startsWith('2.') || !contentLanguage.version) &&
```
glowcloud: Will the first condition not be enough?

```suggestion
    contentLanguage.version?.startsWith('2.') &&
```

packages/apidom-ns-asyncapi-2/src/predicates.ts
```diff
@@ -28,7 +28,7 @@ import ServerVariableElement from './elements/ServerVariable.ts';
 export const isAsyncApi2Element = createPredicate(
   ({ hasBasicElementProps, isElementType, primitiveEq, hasClass }) => {
     return (element: unknown): element is AsyncApi2Element =>
-      element instanceof AsyncApi2Element ||
+      (element instanceof AsyncApi2Element && element.constructor === AsyncApi2Element) ||
```
glowcloud, line 31: I see that other predicates for the root element do not have this - is there some issue when it's not here?
```suggestion
      (element instanceof AsyncApi2Element ||
```

packages/apidom-reference/src/dereference/strategies/asyncapi-3/visitor.ts
```diff
@@ -0,0 +1,535 @@
...
+     */
+    link.replaceWith(mergedElement, mutationReplacer);
+
+    /**
+     * We're at the root of the tree, so we're just replacing the entire tree.
+     */
+    return !parent ? mergedElement : false;
+  }
+
+  public async ChannelItemElement(
```
glowcloud: I think we will never visit here because ChannelItemElement does not exist in AsyncAPI 3.x. We also don't need additional handling for ChannelElement - it was needed for ChannelItemElement because it had the [$ref field](https://v2.asyncapi.com/docs/reference/specification/v2.6.0#channelItemObject), while it [does not exist](https://www.asyncapi.com/docs/reference/specification/v3.0.0#channelObject.) for ChannelElement. As such, I think this can be removed.

packages/apidom-reference/src/dereference/strategies/asyncapi-3/visitor.ts
```diff
@@ -0,0 +1,535 @@
...
+    if (this.options.dereference.immutable) {
+      // register new immutable reference with a refSet
+      const immutableReference = new Reference({
+        uri: `immutable://${baseURI}`,
+        value: parseResult,
+        depth: this.reference.depth + 1,
+      });
+      refSet.add(immutableReference);
+    }
+    // eslint-disable-next-line no-debugger
```
glowcloud:
```suggestion
```

packages/apidom-ls/src/parser-factory.ts
lukaszzazulak: unfortunately there were tests failing without additional condition

packages/apidom-ns-asyncapi-2/src/predicates.ts
lukaszzazulak: yes, the method returns true for the AsyncApi3 as well which is unwanted.

packages/apidom-ls/src/parser-factory.ts
```diff
@@ -48,7 +55,24 @@ export async function parse(
     };
 
     result = await asyncapi2AdapterJson.parse(text, options);
-  } else if (contentLanguage.namespace === 'asyncapi' && contentLanguage.format === 'YAML') {
+  } else if (
```
glowcloud, line 58: Let's move it to be before the YAML adapter, so that they are next to each other, like the other parser adapters.

packages/apidom-ns-asyncapi-2/src/predicates.ts
glowcloud: Right, because of inheritance. Then this seems fine.

packages/apidom-ls/src/parser-factory.ts
glowcloud: It seems to me like the `detect` tests fail then - perhaps we should adjust the [default content language](https://github.com/swagger-api/apidom/blob/feat/async-api3-resolve-references/packages/apidom-ls/test/detect.ts#L37) that's being passed down to the function instead, as it did not provide the version before? 

But not sure if there's other places (e.g. editor) where we would expect to detect Async 2 when there's no version provided.

glowcloud (approved):

packages/apidom-ls/src/parser-factory.ts
robert-hebel-sb: Imo we shouldn't assume spec version if there is no provided

packages/apidom-ls/test/hover-provider.ts
```diff
@@ -468,7 +468,9 @@ describe('apidom-ls-hover-provider', function () {
     }
   });
 
-  it('test hover full provider', async function () {
+  // TODO: Flaky test.
```
robert-hebel-sb, line 471: imo. we should create an issue and attach link here to not forget about fixing it

lukaszzazulak: created OSS-392

packages/apidom-ls/src/parser-factory.ts
lukaszzazulak: ok, changed.

glowcloud (approved):

robert-hebel-sb (approved):

PR #5055, opened 2025-11-24
perf(ls): add cache for getting linting rules

glowcloud:
Adds cache for `getLintingRulesSemantic` to improve validation performance.


packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -152,10 +160,12 @@ export class DefaultValidationService implements ValidationService {
     // get namespace rules with `given` populated as array
     try {
       if (!this.settings?.metadata?.rules) {
```
lukaszzazulak: here direct access, but below assignment to const, maybe const could be moved to the beginning of the try for clarity

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -179,6 +187,7 @@ export class DefaultValidationService implements ValidationService {
     } catch (e) {
       console.log('error in retrieving semantic rules', e);
     }
+    this.lintingRulesSemanticCache.set(cacheKey, meta);
```
lukaszzazulak: it looks like we could set cache in one place only and have same outcome, after calling the getLintingRulesSemantic here 
          const semanticLintingRules = this.getLintingRulesSemantic(api, s, docNs);

PR #5056, opened 2025-11-25
feat(apidom-reference): parser for unit tests

lukaszzazulak:
Parser for resolving references unit tests


packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -454,7 +454,7 @@ export class DefaultValidationService implements ValidationService {
     if (
       nameSpace.namespace === 'asyncapi' &&
       nameSpace.version &&
-      ['3.0.0', '3.0.1'].includes(nameSpace.version)
+      ['3.0.0'].includes(nameSpace.version)
```
glowcloud, line 464: Do we also need to update mediaTypes in namespace https://github.com/swagger-api/apidom/blob/9145c3090023cb5970d55d224037e49a64a8b194/packages/apidom-ns-asyncapi-3/src/media-types.ts#L36 and parser adapters, e.g. README for both [YAML](https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser-adapter-asyncapi-yaml-3#mediatypes) and [JSON](https://github.com/swagger-api/apidom/tree/main/packages/apidom-parser-adapter-asyncapi-json-3#mediatypes) ?

packages/apidom-reference/src/configuration/saturated.ts
```diff
@@ -38,7 +40,9 @@ options.parse.parsers = [
   new OpenAPIJSON3_1Parser({ allowEmpty: true, sourceMap: false }),
   new OpenAPIYAML3_1Parser({ allowEmpty: true, sourceMap: false }),
   new AsyncAPIJSON2Parser({ allowEmpty: true, sourceMap: false }),
+  new AsyncAPIJSON3Parser({ allowEmpty: true, sourceMap: false }),
```
glowcloud: Let's order them with the same convention as the other parsers, so JSON/YAML parsers for the same specification + version are one after another. Let's also amend this ordering in other places where it occurs (README, package.json)
```ts
new AsyncAPIJSON2Parser({ allowEmpty: true, sourceMap: false }),
new AsyncAPIYAML2Parser({ allowEmpty: true, sourceMap: false }),
new AsyncAPIJSON3Parser({ allowEmpty: true, sourceMap: false }),
new AsyncAPIYAML3Parser({ allowEmpty: true, sourceMap: false }),
```

packages/apidom-ls/src/services/validation/validation-service.ts
lukaszzazulak: done.

packages/apidom-reference/src/configuration/saturated.ts
lukaszzazulak: done.

robert-hebel-sb (approved):

PR #5058, opened 2025-12-01
feat(apidom-ns-asyncapi-3): disable resolving references

lukaszzazulak:
Disable resolving references for certain elements


glowcloud (approved):
This seems fine for now before we further decide how to handle these. Just a note that in the future, if we are still not resolving these, we will probably need to find a different solution because of validation of these fields and references.

PR #5062, opened 2025-12-03
test(apidom-ls): fix flaky tests

lukaszzazulak:
Fix flaky tests.


packages/apidom-ls/test/hover-provider.ts
```diff
@@ -438,9 +438,8 @@ describe('apidom-ls-hover-provider', function () {
     logLevel,
   };
 
-  // TODO: Flaky test.
-  // eslint-disable-next-line mocha/no-skipped-tests
-  xit('test hover ref provider', async function () {
+  it('test hover ref provider', async function () {
+    this.timeout(10000);
```
robert-hebel-sb, line 442: this would make tests process longer by 20seconds overall. If we aim for quick fix I would use https://jestjs.io/docs/jest-object#jestretrytimesnumretries-options instead

lukaszzazulak: it will not make it always 20seconds longer; here just sometimes the test exceeds the default timeout 2sec

robert-hebel-sb (approved):

PR #5070, opened 2025-12-15
feat: add option to continue on error when validating references

glowcloud:
Adds a new option for validating references, which will allow the entire definition to be dereferenced, instead of dereferencing each reference separately. The errors will be collected during dereferencing, instead of throwing an error when encountering the first invalid reference.

NOTE 1: We need to dive deep with the visitor to propagate errors from referenced elements so that we replicate the previous validation behaviour. During testing, this did not cause additional issues with performance but I think it sometimes could. Due to this, we could try solving this propagation in a different way, if anyone has any ideas. Overall on tested definitions the reference validation was better by 0,5 - 2,5 seconds, depending on the definition. 

NOTE 2: In OpenAPI 3.1, we additionally need to keep the ancestor referencing element (`ref-referencing-element` meta) so that we can propagate the error from sibling elements, because references can now be defined with other properties in the same element, e.g. how it currently works:
- 3.1:
Screenshot 2025-12-17 at 12 43 40

- 3.0:
Screenshot 2025-12-17 at 12 43 22


cka121 (approved):

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -433,6 +455,117 @@ export class DefaultValidationService implements ValidationService {
...
+              `Reference Error - ${message}`,
+              DiagnosticSeverity.Error,
+              code,
+              'apilint',
+            );
+
+            diagnostic.source = 'apilint';
+            diagnostic.data = {
+              quickFix: [],
+            } as LinterMetaData;
```
robert-hebel-sb: nit:

```suggestion
            } satisfies LinterMetaData;
```
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html
https://dev.to/benarambide/understanding-typescripts-satisfies-vs-as-3d6

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -433,6 +455,117 @@ export class DefaultValidationService implements ValidationService {
...
+            for (const p of pointers) {
+              // @ts-ignore
```
robert-hebel-sb: is it needed here?
```suggestion
```

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -433,6 +455,117 @@ export class DefaultValidationService implements ValidationService {
...
+        );
+        if (message) {
+          const refElement = dereferenceError.refEl;
+          if (refElement as Element) {
+            const refValueElement = (refElement as ObjectElement).get('$ref');
+            const referencedElement = toValue(refElement.getMetaProperty('referenced-element', ''));
+            let pointers = pointersMap[referencedElement];
+            if (!pointers) {
+              pointers = localReferencePointers(doc, referencedElement, true);
+              // eslint-disable-next-line no-param-reassign
```
robert-hebel-sb: removing this line doesn't cause an eslint error
```suggestion
```

PR #5075, opened 2025-12-17
test(reference): add dereference tests for External Docs in AsyncAPI 3 

robert-hebel-sb:
### Description
Tests fails as dereferencing for external docs is broken for schema object. Fix done in https://github.com/swagger-api/apidom/pull/5078

### DIFF
--- a/packages/apidom-reference/test/dereference/strategies/asyncapi-3/server-bindings-object/fixtures/components-server-bindings/dereferenced.json
+++ b/packages/apidom-reference/test/dereference/strategies/asyncapi-3/server-bindings-object/fixtures/components-server-bindings/dereferenced.json
@@ -1,0 +1,15 @@
+[
+  {
+    "asyncapi": "3.0.0",
+    "components": {
+      "serverBindings": {
+        "serverBinding1": {
+          "kafka": {}
+        },
+        "serverBinding2": {
+          "kafka": {}
+        }
+      }
+    }
+  }
+]
--- a/packages/apidom-reference/test/dereference/strategies/asyncapi-3/server-bindings-object/fixtures/components-server-bindings/root.json
+++ b/packages/apidom-reference/test/dereference/strategies/asyncapi-3/server-bindings-object/fixtures/components-server-bindings/root.json
@@ -1,0 +1,13 @@
+{
+  "asyncapi": "3.0.0",
+  "components": {
+    "serverBindings": {
+      "serverBinding1": {
+        "kafka": {}
+      },
+      "serverBinding2": {
+        "$ref": "#/components/serverBindings/serverBinding1"
+      }
+    }
+  }
+}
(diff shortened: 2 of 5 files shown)


packages/apidom-reference/test/dereference/strategies/asyncapi-3/external-documentation-object/fixtures/tags-object/root.json
```diff
@@ -0,0 +1,23 @@
+{
+  "asyncapi": "3.0.0",
+  "components": {
+    "schemas": {
+      "exampleSchema": {
+        "tags": [
```
glowcloud: The `tags` field doesn't exist for schema, from what I'm seeing. The dereferenced file also doesn't have dereferenced `externalDocs` and it would probably not pass if updated.

packages/apidom-reference/test/dereference/strategies/asyncapi-3/external-documentation-object/fixtures/channel-object/root.json
```diff
@@ -0,0 +1,32 @@
...
+        "message1": {
+          "payload": {
+            "type": "object"
+          }
+        }
+      },
+      "externalDocs": {
+        "$ref": "#/components/externalDocs/externalDoc1"
+      },
+      "$ref": "#/components/channels/channel1"
```
glowcloud: The referenced element here will replace any other property of the channel, so the result will not have the dereferenced `externalDocs`. If you remove this, the channel properties should be:
```json
  {
      "address": "users/signup",
      "messages": {
        "message1": {
          "payload": {
            "type": "object"
          }
        }
      },
      "externalDocs": {
        "url": "https://example.com/docs",
        "description": "External documentation for channel"
      }
  }
```

packages/apidom-reference/test/dereference/strategies/asyncapi-3/external-documentation-object/fixtures/operation-object/root.json
```diff
@@ -0,0 +1,27 @@
+{
+  "asyncapi": "3.0.0",
+  "operations": {
+    "getExample": {
+      "externalDocs": {
+        "$ref": "#/components/externalDocs/commonExternalDocs"
+      },
+      "channel": {
+        "$ref": "#/components/channels/channel1"
```
glowcloud: Just a nitpick that we don't need to test dereferencing this here, but we can leave it.

packages/apidom-reference/test/dereference/strategies/asyncapi-3/external-documentation-object/fixtures/channel-object/root.json
robert-hebel-sb: seems I got fooled by the ide built-in linting 😞

packages/apidom-reference/test/dereference/strategies/asyncapi-3/external-documentation-object/fixtures/operation-object/root.json
```diff
@@ -0,0 +1,21 @@
+[
```
glowcloud: I think you might have accidentally copied over the dereferenced definition here, as we're missing the reference to externalDocs in operation + definition is inside of an array 😅

robert-hebel-sb: 🤦‍♂️

PR #5076, opened 2025-12-17
test(reference): add dereference tests for Message in AsyncAPI 3

glowcloud:


robert-hebel-sb:
don't we mosty test here messages object instead of message?

PR #5078, opened 2025-12-17
fix(ns-asyncapi-3): dereference schema external docs

robert-hebel-sb:


glowcloud:
Let's maybe start the commit message with `fix(ns-asyncapi-3)` as the fix is provided there.

glowcloud:
We could also add `externalDocs` in this [test](https://github.com/swagger-api/apidom/blob/d11d8eedb88f347b2aebcb4faa50b9d96dbb8dce/packages/apidom-ns-asyncapi-3/test/refractor/elements/Schema/index.ts#L107) + update the snapshot

PR #5081, opened 2025-12-18
test(ref): add dereference tests for Channel in AsyncAPI 3

cka121:


packages/apidom-ns-asyncapi-3/src/refractor/visitors/async-api-3/operation/ChannelVisitor.ts
```diff
@@ -33,13 +34,12 @@ class ChannelVisitor extends Mixin(FixedFieldsVisitor, FallbackVisitor) {
 
   ObjectElement(objectElement: ObjectElement) {
     const result = FixedFieldsVisitor.prototype.ObjectElement.call(this, objectElement);
-
-    // mark this ReferenceElement with reference metadata
-    if (isStringElement(this.element.$ref)) {
+    
+    if (isReferenceLikeElement(objectElement)) {
```
glowcloud, line 38: Would it work without these changes? I see that we have this in some places in other namespaces, so I would assume it should be correct.

cka121: Test were failing with the existing code so I had to change it to this.

PR #5087, opened 2025-12-30
test(reference): add dereference tests for Message Bindings in AsyncApi3

robert-hebel-sb:


cka121 (commented):
Will require the test cases for message trait as well.

PR #5090, opened 2025-12-30
test(reference): add dereference test for schema object in AsyncAPI3

cka121:
This PR is for [#OSS-383](https://smartbear.atlassian.net/browse/OSS-383)

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,9 +1,7 @@
 {
-  "extends": [
-    "@commitlint/config-conventional"
-  ],
+  "extends": ["@commitlint/config-conventional"],
   "rules": {
-    "header-max-length": [2, "always", 69],
+    "header-max-length": [2, "always", 75],
     "scope-case": [
       2,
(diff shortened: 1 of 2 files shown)


packages/apidom-reference/test/dereference/strategies/asyncapi-3/schema-object/fixtures/message-object/root.json
```diff
@@ -0,0 +1,17 @@
...
+  "asyncapi": "2.6.0",
+  "components": {
+    "schemas": {
+      "schema1": {
+        "title": "schema 1"
+      }
+    },
+    "messages": {
+      "message1": {
+        "headers": {
```
robert-hebel-sb, line 14: could we also test dereferencing for payload property?

cka121: okay

PR #5095, opened 2026-01-12
feat(apidom-ls): docs & config of compatible properties in AsyncAPI3

cka121:
This PR is for [OSS-250](https://smartbear.atlassian.net/browse/OSS-254) and [OSS-251](https://smartbear.atlassian.net/browse/OSS-251)
Note: 
Channel, Operations and Messages will be handled in [OSS-245](https://smartbear.atlassian.net/browse/OSS-245)

### My PR contains...

### My changes...
- [ ] are not breaking changes.

### Documentation

### Automated tests
- [ ] All new and existing tests passed.

### DIFF
--- a/packages/apidom-ls/src/config/common/schema/documentation.ts
+++ b/packages/apidom-ls/src/config/common/schema/documentation.ts
@@ -140,7 +140,7 @@ const documentation = [
   {
     target: 'properties',
     docs: 'The value of "properties" **MUST** be an object. Each value of this object **MUST** be a valid JSON Schema.\n\n  ----  \n\nThis keyword determines how child instances validate for objects, and does not directly validate the immediate instance itself.\n\n  ----  \n\nValidation succeeds if, for each name that appears in both the instance and as a name within this keyword\'s value, the child instance for that name successfully validates against the corresponding schema.\n\n  ----  \n\nOmitting this keyword has the same behavior as an empty object.',
-    targetSpecs: [...AsyncAPI2, ...OpenAPI2, ...OpenAPI3],
+    targetSpecs: [...AsyncAPI2, ...OpenAPI2, ...OpenAPI3, ...AsyncAPI3],
   },
   {
     target: 'patternProperties',
@@ -170,7 +170,7 @@ const documentation = [
   {
     target: 'then',
     docs: 'This keyword\'s value **MUST** be a valid JSON Schema.\n\n  ----  \n\nWhen "if" is present, and the instance successfully validates against its subschema, then validation succeeds against this keyword if the instance also successfully validates against this keyword\'s subschema.\n\n  ----  \n\nThis keyword has no effect when "if" is absent, or when the instance fails to validate against its subschema.  Implementations **MUST** NOT evaluate the instance against this keyword, for either validation or annotation collection purposes, in such cases.',
-    targetSpecs: [...AsyncAPI2, ...OpenAPI3],
+    targetSpecs: [...AsyncAPI2, ...OpenAPI3, ...AsyncAPI3],
   },
   {
     target: 'else',
@@ -205,7 +205,8 @@ const documentation = [
   {
     target: 'contentMediaType',
     docs: 'The value of this property **must** be a media type, as defined by <xref target="RFC2046">RFC 2046</xref>. This property defines the media type of instances which this schema defines.\n\n  ----  \n\nThe value of this property **MUST** be a string.\n\n  ----  \n\nThe value of this property **SHOULD** be ignored if the instance described is not a string.\n\n  ----  \n\nIf the "contentEncoding" property is not present, but the instance value is a string, then the value of this property **SHOULD** specify a text document type, and the character set **SHOULD** be the character set into which the JSON string value was decoded (for which the default is Unicode).',
-    targetSpecs: [...AsyncAPI2, ...OpenAPI3], ...AsyncAPI3,
+    targetSpecs: [...AsyncAPI2, ...OpenAPI3],
+    ...AsyncAPI3,
   },
   {
     target: 'example',


packages/apidom-ls/src/config/asyncapi/message-trait/lint/index.ts
```diff
@@ -18,8 +21,11 @@ import examplesItemsTypeLint from './examples--items-type.ts';
 const lints = [
   allowedFields2_0__2_3Lint,
   allowedFields2_4__2_6Lint,
+  allowedFields3_0Lint,
   messageIdType,
   headersTypeLint,
+  headers3_0TypeLint,
+  payloadTypeLint,
```
glowcloud: There's no `payload` field in Message Trait, it was only added for Message
```suggestion
```

packages/apidom-ls/src/config/asyncapi/message-trait/lint/index.ts
```diff
@@ -18,8 +21,11 @@ import examplesItemsTypeLint from './examples--items-type.ts';
 const lints = [
   allowedFields2_0__2_3Lint,
   allowedFields2_4__2_6Lint,
+  allowedFields3_0Lint,
   messageIdType,
   headersTypeLint,
+  headers3_0TypeLint,
```
glowcloud: Let's use the same naming conventions as in Message, so `headersType3_0Lint` and for the file name `headers--type-3-0.ts`.

packages/apidom-ls/src/config/asyncapi/message-trait/lint/headers-3-0--type.ts
```diff
@@ -0,0 +1,8 @@
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const headers3_0TypeLint: LinterMeta = {
```
glowcloud:
```suggestion
const headersType3_0Lint: LinterMeta = {
```

packages/apidom-ls/src/config/asyncapi/message-trait/lint/allowed-fields-3-0.ts
```diff
@@ -0,0 +1,9 @@
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const allowedFieldsLint3_0: LinterMeta = {
```
glowcloud:
```suggestion
const allowedFields3_0Lint: LinterMeta = {
```

packages/apidom-ls/src/config/asyncapi/info/lint/allowed-fields-3-0.ts
```diff
@@ -0,0 +1,8 @@
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const allowedFieldsLint3_0: LinterMeta = {
```
glowcloud:
```suggestion
const allowedFields3_0Lint: LinterMeta = {
```

packages/apidom-ls/src/config/asyncapi/info/lint/index.ts
```diff
@@ -18,6 +21,9 @@ const lints = [
   contactTypeLint,
   licenseTypeLint,
   allowedFieldsLint,
+  allowedFieldsLint3_0,
```
glowcloud:
```suggestion
  allowedFields3_0Lint,
```

packages/apidom-ls/src/config/asyncapi/components/lint/allowed-fields-3-0.ts
```diff
@@ -0,0 +1,9 @@
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const allowedFieldsLint3_0: LinterMeta = {
```
glowcloud:
```suggestion
const allowedFields3_0Lint: LinterMeta = {
```

packages/apidom-ls/src/config/asyncapi/channel-bindings/lint/allowed-fields-3-0.ts
```diff
@@ -0,0 +1,9 @@
+import { LinterMeta } from '../../../../apidom-language-types.ts';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const allowedFields3_0Lint: LinterMeta = {
```
glowcloud: I think we don't need to add a new allowed fields rule for Channel Bindings, Message Bindings, Operation Bindings and Server Bindings in Async 3, as the fields look to be the same to me as in Async 2.6. We just need to extend the existing rule for 2.6, which is being done [here](https://github.com/swagger-api/apidom/pull/5096/files#diff-e05cf4630ba623d564506d325ba15e6094b428fb2f83a1ab6bb746e75f711494). If I'm wrong about this, then leave it as it is and let me know to revert the change in my PR.

packages/apidom-ls/src/config/asyncapi/async-api-version/documentation.ts
```diff
@@ -1,6 +1,13 @@
...
+    targetSpecs: AsyncAPI2,
+  },
+  {
+    docs: '#### [AsyncAPI Version String](https://www.asyncapi.com/docs/reference/specification/v3.0.0#A2SVersionString)\n\nThe version string signifies the version of the AsyncAPI Specification that the document complies to. The format for this string must be major.minor.patch. The patch may be suffixed by a hyphen and extra alphanumeric characters.\n\\\n\\\nA major.minor shall be used to designate the AsyncAPI Specification version, and will be considered compatible with the AsyncAPI Specification specified by that major.minor version. The patch version will not be considered by tooling, making no distinction between 1.0.0 and 1.0.1.\n\\\n\\\nIn subsequent versions of the AsyncAPI Specification, care will be given such that increments of the minor version should not interfere with operations of tooling developed to a lower minor version. Thus a hypothetical 1.1.0 specification should be usable with tooling designed for 1.0.0.',
```
glowcloud:
```suggestion
    docs: '#### [AsyncAPI Version String](https://www.asyncapi.com/docs/reference/specification/v3.0.0#A2SVersionString)\n\nThe version string signifies the version of the AsyncAPI Specification that the document complies to. The format for this string must be `major`.`minor`.`patch`. The `patch` may be suffixed by a hyphen and extra alphanumeric characters.\n\\\n\\\nA `major`.`minor` shall be used to designate the AsyncAPI Specification version, and will be considered compatible with the AsyncAPI Specification specified by that `major`.`minor` version. The patch version will not be considered by tooling, making no distinction between `1.0.0` and `1.0.1`.\n\\\n\\\nIn subsequent versions of the AsyncAPI Specification, care will be given such that increments of the `minor` version should not interfere with operations of tooling developed to a lower minor version. Thus a hypothetical `1.1.0` specification should be usable with tooling designed for `1.0.0`.',
```

packages/apidom-ls/src/config/asyncapi/parameter/documentation.ts
```diff
@@ -1,22 +1,51 @@
...
+  },
+  {
+    target: 'default',
+    docs: '`string`\n\\\n\\\nThe default value to use for substitution, and to send, if an alternate value is not supplied.',
+    targetSpecs: AsyncAPI3,
+  },
   {
     target: 'description',
     docs: 'A verbose explanation of the parameter. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.',
   },
```
glowcloud, line 18: Need input from others: Should we be adding `targetSpecs: [...AsyncAPI2, ...AsyncAPI3]` to the rules/docs/etc. that apply to both versions? I've been doing that in https://github.com/swagger-api/apidom/pull/5096 to future proof for next spec versions, but maybe it's not necessary / is just nice to have for now? 
Also a note that I think this would not apply to `bindings` - it seems to me that they are meant to be universal. Maybe others also have some knowledge about this, because I'm very unsure about them.

packages/apidom-ls/src/config/asyncapi/bindings/amqp/channel-binding/documentation.ts
```diff
@@ -13,7 +13,7 @@ const documentation = [
   },
   {
     target: 'bindingVersion',
-    docs: 'The version of this binding. If omitted, "0.2.0" MUST be assumed.',
+    docs: 'The version of this binding. If omitted, "0.3.0" MUST be assumed.',
```
glowcloud, line 16: A note here for bindings - it looks to me like the rules and docs might be a bit outdated for some of them. E.g. here the only difference will be the version, but SQS Channel Binding has some fields defined: https://github.com/asyncapi/bindings/tree/master/sqs#channel-binding-object, while the [docs](https://github.com/swagger-api/apidom/blob/main/packages/apidom-ls/src/config/asyncapi/bindings/sqs/channel-binding/documentation.ts) and [linting rules](https://github.com/swagger-api/apidom/blob/main/packages/apidom-ls/src/config/asyncapi/bindings/sqs/channel-binding/lint/latest/allowed-fields.ts) make no mention of them, unless I'm missing something.

Perhaps we should have a separate ticket to go through them and correct them?

packages/apidom-ls/src/config/asyncapi/license/documentation.ts
```diff
@@ -10,5 +10,8 @@ const documentation = [
...
   },
+  {
+    docs: '#### [License Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#licenseObject)\n\nLicense information for the exposed API.\n\n##### Fixed Fields\n\nField Name | Type | Description\n---|:---:|---\nname | `string` | **REQUIRED.** The license name used for the API.\nurl | `string` | A URL to the license used for the API. This MUST be in the form of an absolute URL.\n\nThis object MAY be extended with [Specification Extensions](https://www.asyncapi.com/docs/reference/specification/v3.0.0#specificationExtensions).\n\n##### License Object Example:\n\n\n\\\nJSON\n\n\\\nJSON\n```json\n{\n  "name": "Apache 2.0",\n  "url": "https://www.apache.org/licenses/LICENSE-2.0.html"\n}\n```\n\n\n\\\nYAML\n\n\\\nYAML\n```yaml\nname: Apache 2.0\nurl: https://www.apache.org/licenses/LICENSE-2.0.html\n```',
+  },
```
glowcloud, line 21: Also missing targetSpecs for the Async2 documentation.
```suggestion
  targetSpecs: AsyncAPI3,
  },
```

packages/apidom-ls/src/config/asyncapi/message-trait/documentation.ts
```diff
@@ -9,6 +9,8 @@
  * contain significant information.
  */
 
+import { AsyncAPI2, AsyncAPI3 } from '../target-specs.ts';
+
 const documentation = [
   {
     target: 'messageId',
```
glowcloud, line 16: Seems to me like this is also specific to Async 2

packages/apidom-ls/src/config/asyncapi/message-trait/documentation.ts
```diff
@@ -17,6 +19,12 @@ const documentation = [
...
     docs: '[Schema Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#schemaObject) &#124; [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject) \\\n\\\nSchema definition of the application headers. Schema MUST be of type "object". It **MUST NOT** define the protocol headers.',
+    targetSpecs: AsyncAPI2,
+  },
+  {
+    target: 'headers',
+    docs: '[Multi Format Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject) &#124; [Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#schemaObject) &#124; [Reference Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#referenceObject)\n\\\n\\\nSchema definition of the application headers. Schema MUST be a map of key-value pairs. It **MUST NOT** define the protocol headers. If this is a Schema Object, then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the AsyncAPI Version String.',
+    targetSpecs: AsyncAPI3,
   },
   {
     target: 'payload',
```
glowcloud: This shouldn't be here, AFAIK, but the bug was introduced before. Message Trait documentation says that: This object MAY contain any property from the [Message Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#messageObject), except payload and traits. Same thing for Message Trait in Async 2.

packages/apidom-ls/src/config/asyncapi/message/documentation.ts
```diff
@@ -9,18 +9,33 @@
...
     docs: 'Definition of the message payload. It can be of any type but defaults to [Schema object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#schemaObject). It must match the schema format, including encoding type - e.g Avro should be inlined as either a YAML or JSON object NOT a string to be parsed as YAML or JSON.',
+    targetSpecs: AsyncAPI2,
+  },
+  {
+    target: 'payload',
+    docs: '[Multi Format Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject) &#124; [Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#schemaObject) &#124; [Reference Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#referenceObject)\n\\\n\\\nDefinition of the message payload. If this is a Schema Object, then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the AsyncAPI Version String.',
+    targetSpecs: AsyncAPI3,
   },
   {
     target: 'correlationId',
```
glowcloud, line 41: Should add the one specific to Async 3 as well, with different documentation links, same for `bindings` and `examples` fields?

packages/apidom-ls/src/config/asyncapi/operation-trait/documentation.ts
```diff
@@ -25,17 +28,34 @@ const documentation = [
...
     target: 'bindings',
     docs: '[Operation Bindings Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#operationBindingsObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject)\n\\\n\\\nA map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the operation.',
+    targetSpecs: AsyncAPI2,
+  },
+  {
+    target: 'bindings',
+    docs: '[Operation Bindings Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationBindingsObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#referenceObject)\n\\\n\\\nA map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the operation.',
+    targetSpecs: AsyncAPI2,
   },
   {
```
glowcloud: This might have been added here by accident? It's a previous bug though.

packages/apidom-ls/src/config/asyncapi/operation-trait/documentation.ts
```diff
@@ -9,10 +9,13 @@
  * contain significant information.
  */
 
+import { AsyncAPI2, AsyncAPI3 } from '../target-specs.ts';
+
 const documentation = [
```
glowcloud, line 14: I think we're missing documentation for `title` in Async 3, and `tags` and `externalDocs` for both of them?

But before adding `tags` and `externalDocs` - could you check if it's possible to view them without adding the docs? I think it might be possible for `tags` - similar case is in `info`, where there's no `contact` or `license` documentation, but the hover does show them. I think that if the type of the field is the exact type of that object, then it should show up, so e.g. here `tags` will show up, because the type is `Tags Object`, but `externalDocs` will not, because the type is `External Documentation | Reference Object`. If I'm right, then I'm not sure if we should be adding it or not.

EDIT: I realised that `tags` and `externalDocs` are new keywords, so maybe we don't need to handle it in this PR, although I know you've added them in other places. 😅

packages/apidom-ls/src/config/asyncapi/components/documentation.ts
```diff
@@ -1,62 +1,172 @@
...
+  },
+  {
+    target: 'schemas',
+    docs: 'Map[`string`, [Multi Format Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject) | [Schema Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#schemaObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject)]\n\\\n\\\nAn object to hold reusable Schema Object. If this is a Schema Object, then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the AsyncAPI Version String.    ',
+    targetSpecs: AsyncAPI3,
   },
   {
     target: 'servers',
     docs: 'Map[`string`, [Server Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#serverObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject)]\n\\\n\\\nAn object to hold reusable [Server Objects](https://www.asyncapi.com/docs/reference/specification/v2.6.0#serverObject).',
   },
```
glowcloud, line 18:
```suggestion
  targetSpecs: AsyncAPI2,
  },
```

packages/apidom-ls/src/config/asyncapi/components/documentation.ts
```diff
@@ -1,62 +1,172 @@
+import { AsyncAPI2, AsyncAPI3 } from '../target-specs.ts';
+
 const documentation = [
```
glowcloud, line 3: I think we're missing `operations` for Async 3

EDIT: Similar to `tags` and `externalDocs` in https://github.com/swagger-api/apidom/pull/5095#discussion_r2693381540, I also realised that this is a new keyword, so can leave this for the next PR for non-compatible keywords.

packages/apidom-ls/src/config/asyncapi/bindings/amqp/channel-binding/documentation.ts
cka121: Agreed. We can have a separate ticket to handle this

packages/apidom-ls/src/config/asyncapi/components/documentation.ts
cka121: Yes, I left it on purpose as there is a dedicated ticket for new keywords.

packages/apidom-ls/src/config/asyncapi/operation-trait/documentation.ts
cka121: I deliberately left adding tags and external docs in places where it existed for Async2 as couple of palces have comments written where this tags and externalDocs were ommited on purpose so I let it be as it is and added in places which was available only for Async3 ( I might have missed some places like you spotted).  Should we add it in all these places as well (for Async2 & Async 3 common)? What do you think @glowcloud  ?

packages/apidom-ls/src/config/asyncapi/message-trait/documentation.ts
cka121: Let me recheck this. Thank You!

packages/apidom-ls/src/config/asyncapi/channel-bindings/lint/allowed-fields-3-0.ts
cka121: No I agree with you I just added it for just in case thing> as you confirmed not required I will remove it. Thank You!

packages/apidom-ls/src/config/asyncapi/operation-trait/documentation.ts
glowcloud: `tags` and `externalDocs` documentation will show up even if we don't add it, because it will take it from the element documentation, e.g. this is what I see on https://editor.swagger.io/:

doc_tags
docs_externalDocs


So I think that we can omit adding `tags` but `externalDocs` might be a different case, as it now allows references, so it's `External Documentation | Reference Object`, which should be reflected in the docs - I'm not 100% sure about it though. In general I think that as it changed types, it can be handled in the next PR for non-compatible keywords.

cka121: As disucssed, I will update externalDocs for Async3.

packages/apidom-ls/src/config/asyncapi/components/documentation.ts
```diff
@@ -1,62 +1,173 @@
...
+  },
+  {
+    target: 'schemas',
+    docs: 'Map[`string`, [Multi Format Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject) | [Schema Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#schemaObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject)]\n\\\n\\\nAn object to hold reusable Schema Object. If this is a Schema Object, then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the AsyncAPI Version String.    ',
+    targetSpecs: AsyncAPI3,
   },
   {
     target: 'servers',
     docs: 'Map[`string`, [Server Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#serverObject) | [Reference Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#referenceObject)]\n\\\n\\\nAn object to hold reusable [Server Objects](https://www.asyncapi.com/docs/reference/specification/v2.6.0#serverObject).',
+    targetSpecs: [...AsyncAPI2, ...AsyncAPI3],
```
glowcloud:
```suggestion
    targetSpecs: AsyncAPI2,
```

lukaszzazulak (approved):

PR #5098, opened 2026-01-16
fix(ns-asyncapi-2): add missing reference metadata to server variables

glowcloud:
Adds missing reference metadata in Server Variables Visitor. The missing metadata caused incorrect validation of the `variables` field in Server Object when one of the variables had a `$ref` field.

Before:
variables_before

After:
variables_after

The validation error will correctly show up only if the variable is of invalid type:

variables_after_2

### DIFF
--- a/packages/apidom-ns-asyncapi-2/src/refractor/visitors/async-api-2/server/VariablesVisitor.ts
+++ b/packages/apidom-ns-asyncapi-2/src/refractor/visitors/async-api-2/server/VariablesVisitor.ts
@@ -36,7 +36,7 @@ class VariablesVisitor extends Mixin(MapVisitor, FallbackVisitor) {
   ObjectElement(objectElement: ObjectElement) {
     const result = MapVisitor.prototype.ObjectElement.call(this, objectElement);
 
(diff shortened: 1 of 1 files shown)


packages/apidom-ns-asyncapi-2/src/refractor/visitors/async-api-2/server/VariablesVisitor.ts
```diff
@@ -29,6 +32,17 @@ class VariablesVisitor extends Mixin(MapVisitor, FallbackVisitor) {
         : ['document', 'objects', 'ServerVariable'];
     };
   }
+
+  ObjectElement(objectElement: ObjectElement) {
+    const result = MapVisitor.prototype.ObjectElement.call(this, objectElement);
+
+    // @ts-ignore
```
robert-hebel-sb:
```suggestion
    // @ts-expect-error
```

https://dev.to/maafaishal/ts-expect-error-over-ts-ignore-in-typescript-5140

robert-hebel-sb (approved):

PR #5155, opened 2026-03-31
fix(ls): add validation for examples with date-time format

glowcloud:
Adds OpenAPI validation for `example` and `examples` when format is `date-time`. 

TODO:
- [x] check usage in OpenAPI 2.0 - no need to define it for Parameter Object, as `example` is not defined there
- [x] `null` type and `nullable` property
- [x] day and month associations (e.g. '2025-02-31' should not be valid)
- [x] verify all cases 

Questions:
- should we constrain checking `nullable` to OpenAPI 3.0.x only? 

Notes:
- validating only `example` and `examples` fields in Schema Object, not Example Object


packages/apidom-ls/src/services/validation/linter-functions.ts
```diff
@@ -743,6 +744,91 @@ export const standardLinterfunctions: FunctionItem[] = [
...
+        if (month === 2 && day === 29) {
+          const isLeapYear = (year % 4 === 0 && year % 100 !== 0) || year % 400 === 0;
+
+          if (!isLeapYear) {
+            return false;
+          }
+        }
+
+        return true;
+      };
```
robert-hebel-sb, line 775: There is an upcoming replacement for the JS Date object, which we could use here. It's expected to be released [next year,](https://github.com/tc39/proposals/blob/main/finished-proposals.md) and most browsers support it already; until the support is generally available, we could use this [lightweight polyfill](https://www.npmjs.com/package/temporal-polyfill). What do you think?

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal

````ts
import { Temporal } from 'temporal-polyfill';
.
.
.
const isValidDateTime = (el: Element): boolean => {
  if (!isString(el)) {
    return false;
  }

  const value = toValue(el);

  if (typeof value !== 'string') {
    return false;
  }

  // RFC 3339 requires 'T' (or 't') as the date-time separator;
  // Temporal.Instant.from() also accepts a space (ISO 8601), so we enforce it explicitly.
  if (!/^\d{4}-\d{2}-\d{2}[Tt]/.test(value)) {
    return false;
  }

  try {
    Temporal.Instant.from(value);
    return true;
  } catch {
    return false;
  }
};
````

glowcloud: Nice find! Seems fine to use it and replace it in the future. I added a check to the regular expression to eliminate the calendar/timezone annotations that are allowed in RFC 9557 (`[^[]`).

robert-hebel-sb (approved):

PR #5156, opened 2026-04-01
fix(parser-adapter): move tree lifecycle management to callers

robert-hebel-sb:


packages/apidom-parser-adapter-json/src/lexical-analysis/browser.ts
```diff
@@ -42,11 +40,29 @@ const analyze = async (source: string): Promise<Tree> => {
...
 
-  currentTree = parser.parse(source);
+  // prevent WASM OOM during concurrency spikes by evicting oldest trees
+  // when the pool exceeds threshold; tree.delete() is idempotent so
+  // callers that still hold a reference can safely call delete() again
+  if (activeTrees.size >= MAX_ACTIVE_TREES) {
+    for (const oldTree of activeTrees) {
+      oldTree.delete();
+    }
+    activeTrees.clear();
```
glowcloud: Will calling `oldTree.delete()` for each value not already remove everything from the set?

glowcloud (approved):

packages/apidom-parser-adapter-json/src/lexical-analysis/browser.ts
robert-hebel-sb: good point!

PR #5166, opened 2026-04-27
perf: poc of validation service improvements

glowcloud:
POC of validation improvements:
- YAML direct syntactic analysis with fallback to indirect
- avoid deep visiting when dereferencing with continueOnError option
- AbortSignal handling when starting another validation
- linting functions cache


packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -87,9 +124,11 @@ export class DefaultValidationService implements ValidationService {
 
   private lintingRulesSemanticCache: Map<string, LinterMeta[]> = new Map();
 
-  private referenceNamesCache: string[] = [];
+  private referenceNamesCache: Set<string> = new Set();
 
-  private propertyValuesCache: Map<string, string[]> = new Map();
+  private propertyValuesCache: Map<string, Set<unknown>> = new Map();
+
+  private propertySiblingValuesCache: Map<object, Map<string, Set<unknown>>> = new Map();
```
robert-hebel-sb: could we use WeakMap for these?

packages/apidom-parser-adapter-yaml-1-2/src/syntactic-analysis/direct/index.ts
```diff
@@ -0,0 +1,434 @@
...
+    if (ctx.sourceMap) {
+      member.startPositionRow = pairStartRow;
+      member.startPositionColumn = pairStartCol;
+      member.startIndex = pairStartIdx;
+      member.endPositionRow = pairEndRow;
+      member.endPositionColumn = pairEndCol;
+      member.endIndex = pairEndIdx;
+    }
+    (member as any).content.key = keyElement;
+    (member as any).content.value = valueElement;
```
robert-hebel-sb: perhaps we could use some type guard here 🤔

packages/apidom-ls/src/services/validation/validation-service.ts
```diff
@@ -56,6 +56,43 @@ enum PerfLabels {
...
+    };
+
+    signal.addEventListener('abort', onAbort, { once: true });
+
+    const onFulfilled = (value: T) => {
+      signal.removeEventListener('abort', onAbort);
+      resolve(value);
+    };
+
+    const onRejected = (reason: any) => {
```
robert-hebel-sb:
```suggestion
    const onRejected = (reason: unknown) => {
```

packages/apidom-parser-adapter-yaml-1-2/src/syntactic-analysis/direct/index.ts
```diff
@@ -0,0 +1,434 @@
...
+    const formatted = raw.trimStart().startsWith('>')
+      ? formatBlockFolded(raw)
+      : formatBlockLiteral(raw);
+    const element = ctx.namespace.toElement(formatted);
+    if (ctx.sourceMap) setSourceMap(element, cursor);
+    return element;
+  },
+  walkError: (cursor: Cursor, ctx: DirectContext): null => {
+    // @ts-ignore — currentNode exists on both cursor types
+    const isUnexpected = !(cursor.currentNode as any).hasError;
```
robert-hebel-sb: is it needed?

packages/apidom-parser-adapter-yaml-1-2/src/syntactic-analysis/direct/index.ts
glowcloud: Turns out we can assign directly to key/value. Also, I forgot to commit changes to adapter node/browser so direct analysis was not running 🤦 Fixed this as well.

robert-hebel-sb (approved):

cka121 (approved):

